Issue #63, opened 2017-12-04
Add dependency management and Makefile


hferentschik:
To make it easier to build/test/vet/fmt the code, let's add a Makefile with the appropriate targets. 
Also introduce [https://github.com/golang/dep](dep) as dependency manager.

hferentschik:
Merged via pull request #64

PR #64, opened 2017-12-05
Issue #63 Introducing Makefile, Dep and Travis build

hferentschik:


hferentschik:
@vpavlin - here is the first cut of a Makefile. It contains the standard target to build, test and style check the code. It also contains a target to build the container image. WDYT?

Is there a Docker Hub account we want to push releases/images? In this case, I recommend to also add a release target which builds the container, tags the repo and pushes the image to Docker Hub.

I also enabled Travis CI for continuous integration. See also https://travis-ci.org/hferentschik/fabric8-jenkins-proxy. Once this pull request is merged, we can add the Travis CI integration to the project, so that pull requests are getting build by Travis.

hferentschik:
@vpavlin - once we agree on what we want to be merged,  I will create a equivalent pull request for the idler.

vpavlin:
Thanks @hferentschik! 

Couple notes:
* We do not need Travis, we use Centos CI, I can setup PR builds there
* Release images are pushed to registry.devshift.net (where all openshift.io images go) - https://github.com/fabric8-services/fabric8-jenkins-idler/blob/master/cico_build_deploy.sh#L72
* I'd like to keep using Dockerfile.build if possible as used in https://github.com/fabric8-services/fabric8-jenkins-idler/blob/master/cico_build_deploy.sh#L52-L64 if possible - that helps use as we run on Centos, but we might be developing on a different system (I use Fedora for example)

What do you think?

hferentschik:
> We do not need Travis, we use Centos CI, I can setup PR builds there

+1 for CentOS CI. This sounds like you are already using it? Where is the build script then?

Nothing stops us from having multiple setups. On Minishift - https://github.com/minishift/minishift - we are even using 4, mainly because we need to target all major OS types. The reason I think something like Travis is still valuable is that it gives you a personal CI. Anyone forking this repo and pushing to a branch of this fork can get Travis to build this branch. I find this a very useful feature and often use it before creating a pull request. So my vote would be for CentOS CI + Travis.

We should also update the README with the build status URLs, so that it is easy to find the CI systems.

> Release images are pushed to registry.devshift.net (where all openshift.io images go) - https://github.com/fabric8-services/fabric8-jenkins-idler/blob/master/cico_build_deploy.sh#L72

I see. But in the case of the proxy this is not setup, right? What I am missing is the link to the repo. AFAICT this is just concerned about building and pushing the docker image. IMO a release needs to tag the git repo, build the image from this tag and then push the image to wherever it needs to go.

OOI why doing this in a container. The release steps could be easily added to the Makefile including the mentioned tie in into the GitHub repo.

> I'd like to keep using Dockerfile.build if possible as used in https://github.com/fabric8-services/fabric8-jenkins-idler/blob/master/cico_build_deploy.sh#L52-L64 if possible

Sure. I just noticed I made a change to it. This is atm not required. I will revert this. That said,  I was tempted to remove it. I don't really see its purpose. 

> that helps use as we run on Centos, 

How? The aim would be to let CentOS use the Makefile, right?

> but we might be developing on a different system (I use Fedora for example)

Right, I am on macOS. But the intent of the Makefile is to allow for exactly this. The image build will make sure the right OS is targeted when building the container.

vpavlin:
> Nothing stops us from having multiple setups.

Sounds reasonable


> We should also update the README with the build status URLs,

If you want to add it as a part of this PR, here is the link: https://ci.centos.org/view/Devtools/job/devtools-fabric8-jenkins-idler-build-master/badge/


> But in the case of the proxy this is not setup, right?

Same stuff is setup for Proxy as well (https://ci.centos.org/view/Devtools/job/devtools-fabric8-jenkins-proxy-build-master/)


> IMO a release needs to tag the git repo

We can do it, but what we generally do in fabric8-services is rolling release - every push to master gets built, tested and pushed to our staging environment for OSIO (and potentially promoted to production). If you look at other repos, there are not actual tagged realeases either (https://github.com/fabric8-services/fabric8-tenant, https://github.com/fabric8-services/fabric8-wit)



> I don't really see its purpose.

Multi staged builds are pretty common in container world - they have been added to Dockerfiles even. I agree that it might not make that much sense for Golang as for other languages like Nodejs or Java, but for the sake of consistency, I'd say we should follow the pattern. I think it's fine to have `make build` for local builds and `make image` for CI build where we would use Dockerfile.build and Dockerfile.deploy.

> The aim would be to let CentOS use the Makefile,

Yes, but we run a build on a random Centos machine and then copy the binary to a centos Docker image . I'd prefer to build the binary in a centos docker image and then just move it other centos docker image without the bloat.

hferentschik:
>> Nothing stops us from having multiple setups.

> Sounds reasonable

Ok, let's just go ahead with the current Travis file and we can take it from there (configuring the GitHub config, adding docs, etc)

>> We should also update the README with the build status URLs,

> If you want to add it as a part of this PR, here is the link: https://ci.centos.org/view/Devtools/job/devtools-fabric8-jenkins-idler-build-master/badge/

Ok, I can add the link. That said, I would like to at least have a very short description of what fabric8-jenkins-proxy is. What would be your one paragraph summary?

>> IMO a release needs to tag the git repo

> We can do it, but what we generally do in fabric8-services is rolling release - every push to master gets built, tested and pushed to our staging environment for OSIO (and potentially promoted to production). If you look at other repos, there are not actual tagged realeases either (https://github.com/fabric8-services/fabric8-tenant, https://github.com/fabric8-services/fabric8-wit)

I think this is wrong. I think building and testing on each commit is fine. Also rolling releases. However, the moment an image gets promoted it should be tagged and the underlying code should receive the same tag. How would you be able to rebuild a given image tag/version? This is really a different discussion though.

>> I don't really see its purpose.

> Multi staged builds are pretty common in container world - they have been added to Dockerfiles even. 

Sure, but why even bother with building this in container. With all due respect to containers and what you can do with them, but one does not have to containerize workflows which don't need it. After all it just adds additonal complexity. 

> I agree that it might not make that much sense for Golang 

So why do it?

> as for other languages like Nodejs or Java, but for the sake of consistency,

I am a big fan of consistency, but I don't really think this applies here. After all _fabric8-jenkins-proxy_ stands by itself, right? 

> I think it's fine to have make build for local builds and make image for CI build where we would use Dockerfile.build and Dockerfile.deploy.

The issue I am having with this that it relies on volume mounts. On Linux that might not an issue (even though there one can easily run into them), but once macOS or Windows come into play volume mounts can easily cause permission issues. Often one has to jump through many hoops to make it work. 

I rather keep the code accessible to everyone. In a perfect world you should be able to check it out and work on it, no matter which OS you are using.

> The aim would be to let CentOS use the Makefile,

>> Yes

Cool. The Makefile should contain everything and the CI script calls as much as possible just targets in the Makefile. This way it makes no difference whether you are driving the various steps from the console or from the CI script. It makes it also easier to test the scripts. 

>> but we run a build on a random Centos machine and then copy the binary to a centos Docker image . I'd prefer to build the binary in a centos docker image and then just move it other centos docker image without the bloat.

I am not following you here. Not quite sure what you mean with these various CentOS machines and I don't see the bloat. I even argue you are creating more bloat with a multi stage build.

hferentschik:
Bottom line, I'll add a link to the README and maybe you could provide an overview paragraph we can add as well. 

Then I look at _cico_build_deploy.sh_ and its integration with the _Makefile_. I had missed this part.

vpavlin:
> I am not following you here. Not quite sure what you mean with these various CentOS machines and I don't see the bloat. I even argue you are creating more bloat with a multi stage build.

By bloat I mean golang, git and other dependencies which are needed for build and not for runtime. 

Anyway, seems like you are the one to maintain this for the near future anyway, so up to you - as long as cico_build_deploy.sh produces and pushes image, so that our cluster can deploy, I am fine.


> How would you be able to rebuild a given image tag/version?

The script tags the image with git commit hash, so you can always reproduce by checking out a git ref.



> Bottom line, I'll add a link to the README and maybe you could provide an overview paragraph we can add as well.

I'll add the paragraph once this is merged, ok?

hferentschik:
> By bloat I mean golang, git and other dependencies which are needed for build and not for runtime.

Sure, but they would not end up in the image anyways. Maybe I am missing a piece of the puzzle here, but rest assured my aim is not to the build tools to the final image.

> Anyway, seems like you are the one to maintain this for the near future anyway, so up to you - as long as cico_build_deploy.sh produces and pushes image, so that our cluster can deploy, I am fine.

OK. For each commit/merge to master, that is. It would still be nice to enable pull request builds as well. Any chance you can set them up? This would help to verify the changes I am going to make to the script.

> The script tags the image with git commit hash, so you can always reproduce by checking out a git ref.

Cool. Fair enough then.

> I'll add the paragraph once this is merged, ok?

Works for me.

On a different note, you said:

> every push to master gets built, tested and pushed to our staging environment for OSIO (and potentially promoted to production).

Is this something on top of the CentOS CI builds? Each merge to master gets built and pushed to push.registry.devshift.net via CentOS CI, right? What happens there? Are there any pipelines defined or how else is the promotion happening? Are there any URLs we can add here for reference?

hferentschik:
@vpavlin - what's the meaning of \$CICO_LOCAL? Is this a value injected via CentOS CI? How does it work?

hferentschik:
@vpavlin - I pushed an update with changes to the Makefile and the CentOS build script. I think it has a much clearer separation of concerns. All targets are easily testable. 

Needs some test runs against CentOS CI.

vpavlin:
[test]

vpavlin:
[test]

vpavlin:
> Are there any URLs we can add here for reference?

Here is the master build job: https://ci.centos.org/view/Devtools/job/devtools-fabric8-jenkins-proxy-build-master/

This is how it gets deployed to our staging cluster: https://github.com/almighty/almighty-jobs/blob/master/devtools-ci-index.yaml#L314


> what's the meaning of $CICO_LOCAL? Is this a value injected via CentOS CI? How does it work?

It is simply to allow local testing of the script - i.e. I don't want to install docker and disable selinux locally when I run the script.

Anyway, I've setup PR build in CentOS CI (yay:) )

hferentschik:
@vpavlin, just pushed an update, but it seems there is no rebuild. Not sure whether the GitHub trigger is properly configured in https://github.com/almighty/almighty-jobs/blob/master/devtools-ci-index.yaml. Hard to piece things together with 2300+ lines. 

One would think this template gets applied - https://github.com/almighty/almighty-jobs/blob/master/devtools-ci-index.yaml#L138, but I am not sure.

vpavlin:
[test]

vpavlin:
It is set together, but you need to be in admin list for auto-trigger

hferentschik:
> It is set together, but you need to be in admin list for auto-trigger

Maybe you can add me to whitelist. "add to whitelist"

hferentschik:
@vpavlin pushed again :-)

hferentschik:
retest this please

hferentschik:
[test]

hferentschik:
:-)

PR #86, opened 2017-12-11
Add configuration option for debug mode

vpavlin:


configuration/configuration.go
```diff
@@ -248,3 +249,8 @@ func (c *Data) GetMaxRequestretry() int {
 func (c *Data) GetLocalDevEnv() bool {
 	return c.v.GetBool(varLocalDevEnv)
 }
+
+// GetDebugMode returns if debug mode should be enabled as set via default, config file, or environment variable
+func (c *Data) GetDebugMode() bool {
+	return c.v.GetBool(varDebugMode)
+}
```
hferentschik: To keep GitHub happy one would need a newline

openshift/jenkins-proxy.app.yaml
```diff
@@ -89,6 +89,8 @@ objects:
               secretKeyRef:
                 name: jenkins-proxy
                 key: postgres.password
+          - name: JC_DEBUG
```
hferentschik: What does JC stand for?

vpavlin: When I started to write idler I called it Jenkins Controller, so I used prefix `JC`...would be probably a good thing to refactor in the future so that it makes more sense..

proxy/proxy.go
```diff
@@ -471,7 +471,7 @@ func (p *Proxy) RecordStatistics(ns string, la int64, lbf int64) (err error) {
 		}
 	}
 	if notFound {
-		log.Info("New user %s", ns)
+		log.Infof("New user %s", ns)
```
hferentschik: 👍

configuration/configuration.go
hferentschik: :-)

openshift/jenkins-proxy.app.yaml
hferentschik: +1

hferentschik: I created issue #87

Issue #95, opened 2018-01-10
Remove dependency to fabric8-jenkins-idler


hferentschik:
At the moment there is a dependency to idler in order to reuse the _OpenShift_ and _OpenShiftController_ structs. Instead of this dependency, the router should call the respective REST API endpoint of the Idler.

hferentschik:
Merged via pull request #100

PR #100, opened 2018-01-12
Issue #95 Remove dependency to fabric8-jenkins-idler

vpavlin:
@hferentschik


Gopkg.lock
```diff
@@ -7,11 +7,6 @@
   revision = "dbeaa9332f19a944acb5736b4456cfcc02140e29"
   version = "v3.1.0"
 
-[[projects]]
```
hferentschik: Good to see this dependency go, but you should not touch the lock file. The way to update dependencies is to remove the lock file and then change the dependency Gopkd.toml. Then you rebuild. Dep will detect the missing lock file and re-create it (verifying and resolving dependencies). The you check in and commit the tomal and lock file. Something like this:

    $ make clean && rm Gopkg.lock
    # remove Idler dependency from Gopkd.toml
    $ make all
    $ git add . && git commit

clients/idler.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+
+	log "github.com/sirupsen/logrus"
```
hferentschik: +1

proxy/proxy.go
```diff
@@ -213,13 +209,12 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
 				} else if cookie.Name == CookieJenkinsIdled { //Found a cookie saying Jenkins is idled, verify and act accordingly
 					if c, ok := p.ProxyCache.Get(cookie.Value); ok {
 						pci := c.(ProxyCacheItem)
-						oc := ic.NewOpenShiftWithClient(http.DefaultClient, pci.ClusterURL, pci.OsoToken)
-						isIdle, err := oc.IsIdle(pci.NS, ServiceName)
+						isIdle, err := p.idler.IsIdle(pci.NS)
```
hferentschik: +1. So, for now, we really just need the IsIdle call. Make things imo already clearer. _ProxyCacheItem_ Still has a _OsoToken_. Is that still needed for something? If not we can remove it, together with any helpers (_GetOSOToken_ for example). The less the better and not using the OSO token reduces the amount of places where it is used and hence potentially exposed.

proxy/proxy_cache.go
```diff
@@ -4,14 +4,14 @@ type ProxyCacheItem struct {
 	ClusterURL string
 	NS         string
 	Route      string
-	TLS        bool
+	Scheme     string
 	OsoToken   string
```
hferentschik: I think the OsoToken can be removed now.

proxy/proxy.go
hferentschik: Now I am also more free to continue my refactoring on the Idler. Effectively all becomes _internal_.

Gopkg.lock
vpavlin: Sure, but I haven't touched the lock file:)

proxy/proxy.go
vpavlin: We actually need it to log in to Jenkins, but I need to check if we need to cache it - that might not be needed, let me check

hferentschik: ok

Gopkg.lock
hferentschik: Hmm, one way or another you did. It might be that Dep detected that the dependency to _fabric8-jenkins-idler_ is not needed anymore and updated the lock file (a bit odd). Either way, what we should do is removing the dependency in _Gopkg.toml_, deleting _Gopkg.lock_ and then a rebuild. After that we check in the _toml_ as well as the _lock_ file.

Gopkg.toml
```diff
@@ -19,10 +19,6 @@
 #  name = "github.com/x/y"
 #  version = "2.4.0"
 
-[[constraint]]
```
hferentschik: Rigth, that's what I expected :-)

cico_build_deploy.sh
```diff
@@ -59,7 +59,7 @@ setup_workspace
 
 cd $GOPATH/src/github.com/fabric8-services/fabric8-jenkins-proxy
 echo "HEAD of repository `git rev-parse --short HEAD`"
-make image
+make all
```
hferentschik: :+1 Tests were running on on CentOS CI as well

Makefile
```diff
@@ -5,7 +5,7 @@ REGISTRY_URL = ${REGISTRY_URI}/${REGISTRY_NS}/${REGISTRY_IMAGE}
 IMAGE_TAG ?= $(shell git rev-parse --short HEAD)
 
 BUILD_DIR = out
-PACKAGES = $(shell go list ./...)
+PACKAGES = $(shell go list ./... | grep -v vendor/)
```
hferentschik: We will need this in the Idler Makefile as well

proxy/proxy_cache.go
```diff
@@ -4,16 +4,14 @@ type ProxyCacheItem struct {
 	ClusterURL string
 	NS         string
 	Route      string
-	TLS        bool
-	OsoToken   string
```
hferentschik: +1

hferentschik:
Merged, thanks.

PR #101, opened 2018-01-12
Adding return to prevent 'too many redirects' errors

vpavlin:


proxy/proxy.go
```diff
@@ -354,6 +354,7 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
 				if cached {
 					log.Info(fmt.Sprintf("Redirecting to %s", redirectURL.String()))
 					http.Redirect(w, r, redirectURL.String(), http.StatusFound)
+					return
```
hferentschik: Good catch. I would have loved to see a refactor of this method at the same time. It is just too long and unwieldy. I created https://github.com/fabric8-services/fabric8-jenkins-proxy/issues/103 as a follow up in order to break this method up and reduce the nesting.

Is this change already deployed?

Issue #103, opened 2018-01-12
Refactor Proxy.Handle


hferentschik:
The method is too long and has too many nested if/for statements. In its current form it is hard to track what's going on in this method.

hferentschik:
Merged via pull request #109.

Issue #75, opened 2017-12-08
Create a Replayer component 


hferentschik:
Extract a component for Proxy.ProcessBuffer which acts as "Replayer" (name to be discussed). The idea to refactor Proxy.ProcessBuffer easier to understand (less nested levels).

hferentschik:
See also https://blog.2ndquadrant.com/what-is-select-skip-locked-for-in-postgresql-9-5/

hferentschik:
Relates to #175

PR #109, opened 2018-01-16
Issue #103 Refactor Handle function

vpavlin:
Split out couple blocks of code, happy to do more if you see anything else.

Addresses issue #103

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -104,265 +104,17 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
 		isGH = strings.HasPrefix(ua[0], GHAgent)
 	}
 
-	var body []byte
-	var err error
 	var ns string
 	var cacheKey string
-	var reqURL url.URL
+	var noProxy bool
 	if isGH {
-		//Load request body if it's GH webhok
-		gh := GHHookStruct{}
-		defer r.Body.Close()
-		body, err = ioutil.ReadAll(r.Body)
-		if err != nil {
-			p.HandleError(w, err)
+		ns, noProxy = p.handleGitHubRequest(w, r)
+		if noProxy {
 			return
-		}
-		err = json.Unmarshal(body, &gh)
-		if err != nil {
-			p.HandleError(w, err)
-			return
-		}
-
-		if err != nil {
-			p.HandleError(w, fmt.Errorf("Could not parse GH payload: %s", err))
-			return
-		}
-		log.Info("Processing request from %s", gh.Repository.CloneURL)
-		ns, err = p.GetUser(gh)
-		if err != nil {
-			p.HandleError(w, err)
-			return
-		}
-		scheme, route, err := p.idler.GetRoute(ns)
-		if err != nil {
-			p.HandleError(w, err)
-			return
-		}
-		r.URL.Scheme = scheme
-		r.URL.Host = route
-		r.Host = route
-
-		isIdle, err := p.idler.IsIdle(ns)
-		if err != nil {
-			p.HandleError(w, err)
-			return
-		}
-
-		//If Jenkins is idle, we need to cache the request and return success
-		if isIdle {
-			w.Header().Set("Server", "Webhook-Proxy")
-			sr, err := storage.NewRequest(r, ns, body)
-			if err != nil {
-				p.HandleError(w, err)
-				return
-			}
-			err = p.storageService.CreateRequest(sr)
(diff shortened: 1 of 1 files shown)


proxy/proxy.go
```diff
@@ -172,9 +172,10 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
 			w.WriteHeader(http.StatusAccepted)
 			w.Write([]byte(""))
 			return
-		} else { //If Jenkins is up, we can simply proxy through
-			log.Info(fmt.Sprintf("Passing through %s", r.URL.String()))
 		}
+		//If Jenkins is up, we can simply proxy through
```
hferentschik: I was hoping we could also split here - https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/109/files#diff-3ac192716a1df66bbd1008d327af8792R112, almost splitting the method in half. Something like this:

```golang
   if isGH {
       handleGitHubRequest(w, r)
   } else {
       handleDirectJenkinsUIRequest(w,r)
   }
```

This way the existing comments on this if/else branch become obsolete, since the code gives you the same information.

hferentschik: Similar to the above comment - https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/109/files#diff-3ac192716a1df66bbd1008d327af8792R153

Here we could extract everything into `if isIdle` into a method `sendStatusAcceptedToClient`.

hferentschik: I wonder whether the _redirect_ flag would not be better named as _needsAuth_ or if you want to keep the redirect part _needsAuthRedirect_.

hferentschik: What's about extracting a processCookies and processJSONToken as well?

vpavlin: Thanks! Will look into it

hferentschik:
retest this please

proxy/proxy.go
```diff
@@ -104,296 +104,330 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
...
-		if err != nil {
-			p.HandleError(w, err)
+	} else { //If this is no Github traffic (e.g. user accessing UI)
+		cacheKey, ns, noProxy = p.handleJenkinsUIRequest(w, r)
+		if noProxy {
 			return
 		}
+	}
+	//Write usage stats to DB, run in a goroutine to not slow down the proxy
+	go func() {
```
hferentschik: This we want to refactor, but we can do this later. There are #75 and #115

proxy/proxy.go
```diff
@@ -104,296 +104,330 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
...
+		p.HandleError(w, err)
+		return
+	}
+
+	//If the user provides OSO token, we can directly proxy
+	if _, ok := r.Header["Authorization"]; ok { //FIXME Do we need this?
+		needsAuth = false
+	}
+
+	if len(r.Cookies()) > 0 { //Check cookies and proxy cache to find user info
```
hferentschik: I think here we could further split up, but we can do this in increments.

proxy/proxy.go
```diff
@@ -104,296 +104,330 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
...
-							http.SetCookie(w, cookie)
-						}
-						return //Do not proceed furter - everything is written to ResponseWriter
-					}
+		if nresp.StatusCode == http.StatusOK {
+			for _, cookie := range nresp.Cookies() {
+				if cookie.Name == CookieJenkinsIdled {
+					continue
 				}
+				http.SetCookie(w, cookie)
```
hferentschik: So just to sort this out in my mind, the cookie is just an optimisation which allows for short circuit JWT decoding, tenant lookup, etc. In theory, things would work without the cookie, correct?

hferentschik:
Merged, thanks.

proxy/proxy.go
vpavlin: The flow is 

1. redirect to auth to get JWT token
2. Get user info and OSO token
3. Login to Jenkins with OSO token
4. Find the newly generated JSESSION cookie and pass it back to user - Jenkins uses it to keep the user's session and not to force him re-login for every request, afaiu - which is why we need to tak the cookie from Proxy's login request to Jenkins and pass it back to user so that it gets stored in the browser - as if the user himself logged into Jenkins

Issue #99, opened 2018-01-12
Un-idling Jenkins via directly accessing jenkins.openshift.io results in HTTP 500 error


hferentschik:
> "code": "500",  "detail": "Could not login to Jenkins in manderse-jenkins namespace on behalf of the user manderse@redhat.com"

This happens when trying to directly access _jenkins.openshift.io_ when Jenkins is idled. The Jenkins pod gets scaled up, but then the actual login and forwarding does not work and results in the above error.

If one tries to access _jenkins.openshift.io_ after the pod is up, one gets successfully forwarded.

xcoulon:
I confirm that I get the same error, with the following message:
````
{"Errors":[{"code":"500","detail":"Could not login to Jenkins in xcoulon-jenkins namespace on behalf of the user xcoulon@redhat.com"}]}
````

hferentschik:
Thanks for the feedback. We are in the process of figuring out what's going on there.

hferentschik:
Fixed via pull request #116

PR #116, opened 2018-01-18
Issue #99 Un-idling Jenkins via directly accessing proxy results in …

vpavlin:
…HTTP 500 error

The problem is OpenShift HAProxy routers take a long time to reconfigure and so
even if Jenkins is up and running, the route might not be accessible yet.

Adding a check for the route to continue with waiting if it is not
available

addresses issue #99

### DIFF
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -198,12 +198,16 @@ func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request) (
 					err = p.processTemplate(w, ns)
 					p.RecordStatistics(pci.NS, time.Now().Unix(), 0) //FIXME - maybe do this at the beginning?
 				} else { //If Jenkins is running, remove the cookie
-					var resp *http.Response
 					//OpenShift can take up to couple tens of second to update HAProxy configuration for new route
 					//so even if the pod is up, route might still return 500 - i.e. we need to check the route
 					//before claiming Jenkins is up
-					resp, err = p.loginJenkins(pci, "")
-					if resp.StatusCode == 200 || resp.StatusCode == 403 {
+					var statusCode int
+					statusCode, _, err = p.loginJenkins(pci, "")
+					if err != nil {
+						p.HandleError(w, err)
+						return
+					}
+					if statusCode == 200 || statusCode == 403 {
 						cookie.Expires = time.Unix(0, 0)
 						http.SetCookie(w, cookie)
 					} else {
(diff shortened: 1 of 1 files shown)


vpavlin:
Rebased

internal/proxy/proxy.go
```diff
@@ -108,6 +108,9 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
 	var ns string
 	var cacheKey string
 	var noProxy bool
+
```
hferentschik: +1

internal/proxy/proxy.go
```diff
@@ -195,8 +198,17 @@ func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request) (
 					err = p.processTemplate(w, ns)
 					p.RecordStatistics(pci.NS, time.Now().Unix(), 0) //FIXME - maybe do this at the beginning?
 				} else { //If Jenkins is running, remove the cookie
-					cookie.Expires = time.Unix(0, 0)
-					http.SetCookie(w, cookie)
+					var resp *http.Response
```
hferentschik: AFAIK you need to close the response body. Check the godoc or see here https://stackoverflow.com/questions/33238518/what-could-happen-if-i-dont-close-response-body-in-golang.

You should be able to use _defer_ in _loginJenkins_.

That said, do you really need to return the full reponse? Could you not just return status code and error?

internal/proxy/proxy.go
```diff
@@ -195,8 +198,17 @@ func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request) (
...
+					//OpenShift can take up to couple tens of second to update HAProxy configuration for new route
+					//so even if the pod is up, route might still return 500 - i.e. we need to check the route
+					//before claiming Jenkins is up
+					resp, err = p.loginJenkins(pci, "")
+					if resp.StatusCode == 200 || resp.StatusCode == 403 {
```
hferentschik: This might panic. You are effectively ignoring the error here. If there is an underlying error, the response will be nil in which case the call to _resp.StatusCode_ will panic. Usually in Go if a method returns an error, you always have to first check the error before doing anything else with any other returned object. More often than not, if an error occurs other return parameters are nil.

internal/proxy/proxy.go
vpavlin: Good catch, I'll fix it

The problem with just returning the code is that I used the same function here https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/116/files#diff-80a67e0d6b63df2dd084ee5a102f2f62R262 where we need cookies, so we'd have to return statusCode, cookies and err - I can do that if you think it's better

internal/proxy/proxy.go
vpavlin: Yeah, sorry, I over optimized there - leaving the error check for later, but obviously resp == nil would cause panic, thanks

vpavlin:
@hferentschik I've update the PR...no longer returning whole response.

Also wanted to refactor the naked return in the long function, but then realised it would be better to do it in separate PR.

If you could re-review, please

hferentschik:
> I've updated the PR...no longer returning whole response.

cool

> Also wanted to refactor the naked return in the long function, but then realised it would be better to do it in separate PR.

+1

hferentschik:
Merged (squashed commits)

Issue #108, opened 2018-01-15
Create performance test harness


hferentschik:
Before enabling Proxy for all users, we need to run some basic performance validation. One thing we can do is to create GitHub webbhook requests and see how the system behaves. Another thing we can do is to simulate some "more than normal" usage by a selected number of users.

hferentschik:
First cut merged via pull request #117

PR #117, opened 2018-01-22
[WIP] Issue #108  Adding siegeJenkins.sh to battle test Jenkins Proxy

hferentschik:
- Creating osio helper tool to create JSON token from Red Hat Developer credentials
- Creating Siege setup to load test proxy


cmd/osio/cmd/token.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func init() {
+	cmdToken.Flags().StringVarP(&targetEnv, "target", "t", "stage", "Target environment OpenShift.io stage vs prod.")
+	cmdToken.Flags().StringVarP(&username, "username", "u", "", "Red Hat Developer username.")
+	cmdToken.Flags().StringVarP(&password, "password", "p", "", "Red Hat Developer password.")
+	cmdToken.Flags().BoolVarP(&encode, "encode", "e", false, "Whether or not the output should be URL encoded.")
+}
+
+func runCreateToken(cmd *cobra.Command, args []string) {
```
hferentschik: @alexeykazakov, FYI, this is the helper tool I ended up creating as part of _fabric8-jenkins-proxy_. Given Red Hat Developers credentials it spits out the JSON access token. I would like to extend this to also create OpenShift access token and maybe some other things. My idea would be that this tool can help in setting up scripted access to the system, for example for testing purposes. 

Benefit over the offline token is that this does not expire and I don't have any manual steps. That said, osio could also be used to create the offline token. Same same really.

internal/util/token.go
```diff
@@ -0,0 +1,124 @@
...
+	environments = make(map[string]environment, 2)
+)
+
+func init() {
+	environments[stage] = environment{
+		authURL:     "https://auth.prod-preview.openshift.io/api/login",
+		redirectURL: "https://jenkins.prod-preview.openshift.io",
+	}
+
+	environments[prod] = environment{
```
vpavlin: I don't like harcoding these URLs in the git repo

vpavlin:
Generally looks good to me, one concern is the hardcoding of prod(-prev) URLs - I'd like to avoid that as much as possible.

Second thing is - so far it seems like a very general tool, wouldn't it make sense to have this in a separate repository?

internal/util/token.go
hferentschik: What's the issue? It is not really any secret URL, is it? And more importantly, what's the alternative? To pass the URLs as command line arguments. To a certain degree, it will diminish the value/convenience of this tool.

hferentschik:
> Generally looks good to me, one concern is the hardcoding of prod(-prev) URLs - I'd like to avoid that as much as possible.

What's the issue and what's the alternative?

> Second thing is - so far it seems like a very general tool, wouldn't it make sense to have this in a separate repository?

Yes and no. Personally, I see potential here. I can see that _osio_ develops as some sort of swiss army knife around small functions one needs when dealing with OpenShift.io. This could be related to auth, but also other things. Right now we just have one single function implemented. Whether this ever happens depends on there is any interest/need. I am a believer and letting these things grow organically.  It is kind of convenient to have it in this repo and it is easy to extract into its own repo when/if there is interest.

vpavlin:
> What's the issue and what's the alternative?

You are right, it's not a secret, but those are "product" URLs, but this is an upstream repository, so why should be hardcode these URLs in the code?

If someone wants to contribute to fabric8, they'll run it on their cluster with their own URLs - why should default go to prod instances by default?

Command line arguments, ENV vars or a simple config file - all sound better to me than hardcoding any DNS in the code

hferentschik:
> You are right, it's not a secret, but those are "product" URLs, but this is an upstream repository, so why should be hardcode these URLs in the code?

This is really a very specific tool which only makes sense for our "products". If you were to run your own cluster/instance you would not have any use of this tool, since you very likely would not authenticate against Red Hat Developers. Even if we externalize these URLs the valid URLs would only be limited to the ones used here.

hferentschik:
I think we could merge this now as a first cut.

hferentschik:
Cheers.

Issue #81, opened 2017-12-11
Put storage service behind an interface


hferentschik:

hferentschik:
Merged via pull request #118

PR #118, opened 2018-01-23
Issue #81 Creating Store interface to abstract to data layer

hferentschik:
- Refactored main.go to use WaitGroup and Context to start goroutines
 - Introducing testify and dockertest.v3 as test dependencies
 - Removing testrun.go

Addressing issue #81
Based on pull request #117


hferentschik:
@vpavlin Have a look if you can. I need to make a couple of minor changes still, but I'd like to get this in asap. This refactors the storage part (introducing an interface :-)) as well as starting to handle the various goroutines in a better way.

cmd/fabric8-jenkins-proxy/main_test.go
```diff
@@ -0,0 +1,93 @@
...
+	code := m.Run()
+
+	// When you're done, kill and remove the container
+	err = pool.Purge(resource)
+
+	os.Exit(code)
+}
+
+func TestProxy(t *testing.T) {
+	t.SkipNow()
```
hferentschik: @vpavlin This one intrigues me. This test is supposed to be a replacement for _testrun.go_. Among other things, this test will actually create and start the test DB as part of the test. 
When I run this, I get:

    {"level":"fatal","msg":"Get /auth/realms/fabric8/: unsupported protocol scheme \"\"","time":"2018-01-23T13:45:30+01:00"}

I don't think I broke this. Any ideas? Shall we try to get this working end to end again? If you want to try, get the branch and uncomment this line and run the tests.

cmd/fabric8-jenkins-proxy/main.go
```diff
@@ -50,48 +58,94 @@ func main() {
...
+	var wg sync.WaitGroup
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
 
-	//Listen for API
+	startWorkers(wg, ctx, cancel, store, &proxy, defaultStatsLoggingInterval)
+	wg.Wait()
+}
+
+func startWorkers(wg sync.WaitGroup, ctx context.Context, cancel context.CancelFunc, store storage.Store, proxy *proxy.Proxy, interval time.Duration) {
```
hferentschik: @vpavlin So here we take care of the goroutines now. Not all of them are added yet (the goroutine processing the request buffer is for example still missing). So if you work on a StatsRecorder it should be started here.

vpavlin:
Is the combination with https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/117 intentional?

cmd/fabric8-jenkins-proxy/main_test.go
vpavlin: This https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/118/files#diff-951fb2cd29666d1232c48e2dd695609fR88 needs to resolve to https://github.com/hferentschik/fabric8-jenkins-proxy/blob/293291177020548dd5df52f5295eefd64e8a7207/cmd/fabric8-jenkins-proxy/main_test.go#L69 as in https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/master/internal/testutils/testrun.go#L58

hferentschik:
> Is the combination with #117 intentional?

I was basing it of #117, since I wanted to reduce the amount of times I had to recreate the dep lock file. It is now rebased on master

cmd/fabric8-jenkins-proxy/main_test.go
hferentschik: thanks, I'll look into it.

hferentschik: fixed

hferentschik:
@vpavlin have a look now. I think the actual objective of the issue is met now (and more). There are a couple of TODOs left, but I want to address them as separate issues.

cmd/fabric8-jenkins-proxy/main_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	go func() {
+		// Send SIGTERM after two seconds
+		time.Sleep(2 * time.Second)
+		syscall.Kill(syscall.Getpid(), syscall.SIGTERM)
+	}()
+
+	start(config, &tenant, &wit, &idler, store)
+
+	// TODO - Test an actual workflow by triggering some of the MockURLs
```
hferentschik: @vpavlin So here we go. The proxy comes up and atm gets shut down after two seconds. ATM, I am just asserting that the shutdown happens as expected. Really in between, one should trigger an actual workflow, but tbh I am not sure where things are in terms of everything being up to date. For example, I tried to hit the Jenkins proxy URL and got an error. `JenkinsStatus` in common.go seems to expect a file "code.txt" which does not seem to exist.

Issue #94, opened 2018-01-08
Update GitHub webhook URLs 


hferentschik:
GitHub webhooks triggering Jenkins build have the form https://jenkins-<user\>-jenkins.8a09.starter-us-east-2.openshiftapps.com/github-webhook/

In order for the Jenkins Proxy to work, the webhook URL needs to be changed when the idling is enabled for a given user. 

The proxy should on startup iterate the list users and determine which user has the idler feature enabled. For these users, the webhook URL needs to be checked and potentially updated. 

The idea would be to delegate the change of the URL to the launchpad (generator-backend) which should already have the required code to set/modify the webhook. @gastaldi WDYT?

gastaldi:
I think it's fine, as long as we don't add any specific logic related to other services in the launcher-backend. Is there a service we can call passing the authorization bearer token and the Jenkins webhook URL as in the format you described and return the modified URL (if Idler is enabled for the current user in this case)?

hferentschik:
>  Is there a service we can call passing the authorization bearer token and the Jenkins webhook URL 

Well, I was hoping that you could be this service ;-) After all you have the code to create the webhook in the first place, right?

gastaldi:
Yeah, we can create the webhook, but I'd prefer to keep the specific business logic separated (like checking if the user should use Idler or not) to avoid maintenance nightmares ;)

vpavlin:
My idea would be to have 2 endpoints in backend:

1. Get list of existing webhook URLs

Now our code would check if the one we expect there is present

2. Update the existing webhook URL pointing to Jenkins with Proxy url

That would only mean backend needs to be able to list webhooks and update a single webhook - no logic hidden there

vpavlin:
One note to this approach - this would work for current approach - enable specific users via toggle service, once we enable it for all, we need to switch all users to the proxy anyway, because looping through all of them and doing these checks would not make sense

gastaldi:
@vpavlin isn't the webhook creation supposed to be performed when a quickstart is created? Or is it invoked from Jenkins or other system? We can definitely create the endpoints, I am just not sure who will call it and when

vpavlin:
Yes, the webhook generation is handled by generator backend when codebase is generated, problem is we have existing tenants and we need to migrate them to new Jenkins route (jenkins-$USER-jenkins..openshift.com -> jenkins.openshift.io).

[...]

Issue #114, opened 2018-01-17
Update Jenkins build log URLs from OpenShift console pipeline view


hferentschik:
On the pipeline page w/i the OpenShift console, there are links to the Jenkins build logs. These links refer directly to the user-specific Jenkins instance/route. These links needs to point to the Jenkins Proxy URL, so that in case of an idled Jenkins instance, the service can be first scaled up again.

This issue is a placeholder to find out where code needs to change to make this happen.

hferentschik:
@joshuawilson any update on your investigation regarding what drives the generation of this link and how we can update it?

hferentschik:
@joshuawilson Do you have an update? Do we at least by know what drives these links?

joshuawilson:
We are working this, hope to have it done today. If not then early tomorrow. Need to get past the CORS issue.

hferentschik:
Ok, thanks. Keep us posted.

joshuawilson:
waiting on the SD team to update the URLs, I have the ticket in requesting it

hferentschik:
Mind sharing where the change needs to be done? Do you have al ink to the issue? What's the plan? Can we make this change for prod-preview first and if it looks good we will change prod once we are ready to full rollout Jenkins Proxy?

joshuawilson:
The issue is at https://github.com/openshiftio/openshift.io/issues/1685
There are multiple PRs that change parts in the code to support it. 
This PR sets the URL https://github.com/openshiftio/saas-openshiftio/pull/598
I'm unsure if I can set that for only prod-preview and not break prod but I'm going to try.

hferentschik:
@joshuawilson I think we are talking about different things. AFAIU fabric8-oso-proxy is needed for multi-cluster support. This is not what we are after here. We are talking about the Jenkins Proxy - fabric8-jenkins-proxy - which proxies access to the Jenkins instance in the user's workspace. In order for us to idle the Jenkins instance we need to make sure that all access (direct UI access as well as webhooks) are proxied via fabric8-jenkins-proxy. The webhook part is covered is covered in issue #94 which will update existing GitHub webhooks. 

What we try to achieve here is to make sure that the UI (the page which shows the stages of the pipeline during a build) get updated as well. The pipeline view contains a link to the Jenkins logs. This link/URL needs to change to go through the Jenkins Proxy.

hferentschik:
[...]

PR #123, opened 2018-02-01
Issue #94 Dockerfile and scripts for web hook migration

hferentschik:
@vpavlin - Still work in progress, but have a look and let me know whether I am on the right track. As discussed I am now building a container, so that we can run the migration inside the cluster. 

Works nicely so far from my machine as well. 

Addresses issue #94

### DIFF
--- a/internal/util/osio_token.go
+++ b/internal/util/osio_token.go
@@ -1,0 +1,145 @@
+package util
+
+import (
+	"crypto/rsa"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/dgrijalva/jwt-go"
+	"github.com/satori/go.uuid"
+	"time"
+)
+
+type User struct {
+	Data Data `json:"data"`
+}
+
+type Data struct {
+	Attributes Attributes `json:"attributes"`
+}
+
+type Attributes struct {
+	Username string `json:"username"`
+	Email    string `json:"email"`
+	FullName string `json:"fullName"`
+}
+
+type Result struct {
+	Tokens []TokenResult `json:"tokens"`
+}
+
+type TokenResult struct {
+	UUID  string `json:"uuid"`
+	Token string `json:"token"`
+}
+
+type Token struct {
+	Data Data `json:"data"`
+}
+
+func CreateOSIOToken(env string, uuid string, key string, valid int64, session string) (string, error) {
+	privateKey, err := jwt.ParseRSAPrivateKeyFromPEM([]byte(key))
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("Unable to parse private key file: %s", err))
+	}
+
+	user, err := loadUser(env, uuid)
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("Unable to GET user for uuid '%s': %s", uuid, err))
+	}
+
+	token, err := generateToken(env, privateKey, user, uuid, valid, session)
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("Unable to generate token for uuid '%s': %s", uuid, err))
+	}
+
+	return token, nil
+}
+
+// generateToken generates the JWT token using the specified private key.
+// See also https://self-issued.info/docs/draft-ietf-oauth-json-web-token.html.
+func generateToken(env string, key *rsa.PrivateKey, user User, userID string, valid int64, session string) (string, error) {
+	token := jwt.New(jwt.SigningMethodRS256)
+
+	token.Header["kid"] = environments[env].privateKeyId
+
+	// JWT ID, needs to be unique
+	token.Claims.(jwt.MapClaims)["jti"] = uuid.NewV4().String()
+
+	// Deal with various time related fields
+	now := time.Now().Unix()
+	token.Claims.(jwt.MapClaims)["iat"] = now            // Token issue time
+	token.Claims.(jwt.MapClaims)["exp"] = now + 60*valid // Token expiry time
+
+	token.Claims.(jwt.MapClaims)["iss"] = "https://sso.openshift.io/auth/realms/fabric8"
+	token.Claims.(jwt.MapClaims)["aud"] = "fabric8-online-platform"
+	token.Claims.(jwt.MapClaims)["typ"] = "Bearer"
+	token.Claims.(jwt.MapClaims)["azp"] = "fabric8-online-platform"
+	token.Claims.(jwt.MapClaims)["sub"] = userID
+
+	token.Claims.(jwt.MapClaims)["acr"] = "1"
+	token.Claims.(jwt.MapClaims)["approved"] = "true"
+	token.Claims.(jwt.MapClaims)["company"] = ""
+	token.Claims.(jwt.MapClaims)["name"] = user.Data.Attributes.FullName
+	token.Claims.(jwt.MapClaims)["preferred_username"] = user.Data.Attributes.Username
+	token.Claims.(jwt.MapClaims)["given_name"] = ""
+	token.Claims.(jwt.MapClaims)["family_name"] = ""
+	token.Claims.(jwt.MapClaims)["email"] = user.Data.Attributes.Email
+
+	if len(session) > 0 {
+		token.Claims.(jwt.MapClaims)["session_state"] = session
+	}
+
+	token.Claims.(jwt.MapClaims)["allowed-origins"] = []string{
+		environments[env].osioUrl,
+		environments[env].authURL,
+		environments[env].apiUrl,
+	}
+
+	realmAccess := make(map[string]interface{})
+	realmAccess["roles"] = []string{"uma_authorization"}
+	token.Claims.(jwt.MapClaims)["realm_access"] = realmAccess
+
+	resourceAccess := make(map[string]interface{})
+	broker := make(map[string]interface{})
+	broker["roles"] = []string{"read-token"}
+	resourceAccess["broker"] = broker
+
+	account := make(map[string]interface{})
+	account["roles"] = []string{"manage-account", "manage-account-links", "view-profile"}
+	resourceAccess["account"] = account
+
+	token.Claims.(jwt.MapClaims)["resource_access"] = resourceAccess
+
+	tokenStr, err := token.SignedString(key)
+	if err != nil {
+		return "", err
+	}
+	return tokenStr, nil
+}
+
+func loadUser(env string, uuid string) (User, error) {
+	var user User
+	req, err := http.NewRequest("GET", environments[env].authURL+"/api/users/"+uuid, nil)
+	if err != nil {
+		return user, err
+	}
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return user, err
+	}
+	defer res.Body.Close()
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return user, err
+	}
+	if res.StatusCode != http.StatusOK {
+		return user, errors.New("Status is not 200 OK: " + res.Status)
+	}
+	err = json.Unmarshal(body, &user)
+
+	return user, err
+}
--- a/scripts/webhook-migration/updateWebhooks.sh
+++ b/scripts/webhook-migration/updateWebhooks.sh
@@ -1,0 +1,231 @@
+#!/bin/bash
+#
+# Used to update GitHub webhooks.
+# See https://github.com/fabric8-services/fabric8-jenkins-proxy/issues/94
+
+# Make sure we have all the required environment variables injected
+[ -z "${TENANT_URL}" ] && echo "TENANT_URL needs to be set." && exit 1
+[ -z "${AUTH_URL}" ] && echo "AUTH_URL needs to be set" && exit 1
+[ -z "${OPENSHIFT_URL}" ] && echo "OPENSHIFT_URL needs to be set." && exit 1
+[ -z "${OPENSHIFT_API_TOKEN}" ] && echo "OPENSHIFT_API_TOKEN needs to be set." && exit 1
+[ -z "${AUTH_TOKEN}" ] && echo "AUTH_TOKEN needs to be set." && exit 1
+[ -z "${JENKINS_OLD_URL_SUFFIX}" ] && echo "JENKINS_OLD_URL_SUFFIX needs to be set." && exit 1
+[ -z "${JENKINS_PROXY_URL}" ] && echo "JENKINS_PROXY_URL needs to be set." && exit 1
+[ -z "${PRIVATE_KEY}" ] && echo "PRIVATE_KEY needs to be set." && exit 1
+[ -z "${TARGET_ENV}" ] && echo "TARGET_ENV needs to be set. [prod|stage]." && exit 1
+
+DRY_RUN="${DRY_RUN:-false}"
+
+# Ensure required tools are installed
+hash jq 2>/dev/null || { echo >&2 "'jq' is required."; exit 1; }
+hash curl 2>/dev/null || { echo >&2 "'curl' is required."; exit 1; }
+hash oc 2>/dev/null || { echo >&2 "'oc' is required."; exit 1; }
+hash osio 2>/dev/null || { echo >&2 "'osio' is required."; exit 1; }
+
+###############################################################################
+# Processes the build configuration of
+# the OpenShift cluster and extracts GitHub
+# repo URLs which might need updating.
+# Globals:
+#   OPENSHIFT_URL       - The OpenShift URL containing the user namespaces
+#   OPENSHIFT_API_TOKEN - API token to access OpenShift. Needs to have permission to list all namespaces
+#   CHECK_NAMESPACES    - List of namespaces to check. For testing purposes. If not set all namespaces are iterated
+#   NAME_SPACES         - Array of namespaces which contain a git URL one of their build configurations
+#                         This array is populated as part of the execution of this method.
+#   REPOS               - Array of Git repository URLs aligning by index with NAME_SPACES.
+#                       - This array is populated as part of the execution of this method.
+# Arguments:
+#   None
+# Returns:
+#   None
+###############################################################################
+findGitHubRepos() {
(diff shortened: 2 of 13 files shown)


scripts/webhook-migration/updateWebhooks.sh
```diff
@@ -0,0 +1,137 @@
...
+#   UUIDS      - User ids to iterate
+#   OSIO_TOKEN - Array which gets populated with OSIO tokens
+# Arguments:
+#   None
+# Returns:
+#   None
+#######################################
+getOsioTokens() {
+  for i in "${!UUIDS[@]}"; do
+    # TODO Issue #94 - Here we have to call token-generator.
```
hferentschik: @alexeykazakov @vpavlin - here I need to convert a user's uuid to an OSIO token. Any chance we can change `token-generator` to take a single uuid as input and return the token as output? This way I could easily call `token-generator` in this loop iterating over uuids? 
Or is there a reason why currently token generation in batch?

scripts/webhook-migration/updateWebhooks.sh
```diff
@@ -0,0 +1,137 @@
...
+#   OSIO_TOKENS    - OSIO tokens to iterate
+#   GITHUB_TOKENS  - Array which gets populated with GitHub tokens
+# Arguments:
+#   None
+# Returns:
+#   None
+#######################################
+getGitHubTokens() {
+  for i in "${!OSIO_TOKENS[@]}"; do
+    # TODO Issue #94 - Determine how to get GH token
```
hferentschik: @vpavlin do you know out of the top of the head which API call I need to make to get the GitHub token? If you know the URL and parameters that would be great, then I can plug things in.

scripts/webhook-migration/updateWebhooks.sh
```diff
@@ -0,0 +1,137 @@
...
+#
+# Arguments:
+#   None
+# Returns:
+#   None
+#######################################
+updateGitHubWebhook() {
+  for i in "${!REPOS[@]}"; do
+    # TODO Issue #94 - Update a webhook
+    # https://developer.github.com/v3/repos/hooks
```
hferentschik: @vpavlin I'll work on the actual update of the hook now...

hferentschik:
@vpavlin All pieces in place now. I am not able to test the OSIO token generation. I will not be able to test this until I can run this in the cluster with everything set up.

scripts/webhook-migration/updateWebhooks.sh
```diff
@@ -0,0 +1,182 @@
...
+#   OSIO_TOKEN - Array which gets populated with OSIO tokens
+# Arguments:
+#   None
+# Returns:
+#   None
+###############################################################################
+getOsioTokens() {
+  for i in "${!UUIDS[@]}"; do
+    # TODO Issue #94 - How to pass the key? ATM it is a file path? Should it be read from env?
+    osio_token=$(echo "osio token -k keyfile -u ${UUIDS[$i]}")
```
hferentschik: ATM the `osio` CLI expects a file path to the private key. I guess for the in-container mode, we read the key also out of the environment? Or would it be mounted?

scripts/webhook-migration/updateWebhooks.sh
```diff
@@ -0,0 +1,182 @@
...
+    echo ${osio_token}
+
+    # TODO Issue #94 - This will need to be fixed
+    # MY_SECRET is single injected OSIO token for test purposes
+    OSIO_TOKENS[$i]=$MY_SECRET
```
hferentschik: This is just a hack to inject my token for testing.

hferentschik:
@vpavlin How do we get this container to run in the cluster and inject the required properties? Do we need to add some YAML config files as well?

scripts/webhook-migration/updateWebhooks.sh
vpavlin: The easiest would be to get it from env var

scripts/webhook-migration/updateWebhooks.sh
```diff
@@ -0,0 +1,182 @@
...
+
+    hook_urls=($(curl -sgSL -H "Authorization: token ${GITHUB_TOKENS[$i]}" \
+      https://api.github.com/repos/${REPOS[$i]}/hooks | \
+      jq -r '.[] | .config.url' | tr '\n' ' '))
+
+    # That's a bit flaky since it assumes the order of hooks is always the same
+    # Seems to be this way
+    hook_ids=($(curl -sgSL -H "Authorization: token ${GITHUB_TOKENS[$i]}" \
+      https://api.github.com/repos/${REPOS[$i]}/hooks | \
+      jq -r '.[] | .id' | tr '\n' ' '))
```
vpavlin: How about writing output of first curl to a file or env var and then using it to parse urls and ids? This way we might be more sure about the order and content being consistent

vpavlin: Also, you could probably get the url and ids from the json with jq in one call and then split it in bash (and avoid 2 curl calls)

scripts/webhook-migration/updateWebhooks.sh
```diff
@@ -0,0 +1,182 @@
...
+
+    for i in "${!hook_urls[@]}"; do
+      if [[ ${hook_urls[${i}]} =~ .*${JENKINS_OLD_URL_SUFFIX}$ ]]; then
+        response=$(curl -sgSL -H "Authorization: token ${GITHUB_TOKENS[$i]}" -X PATCH \
+          --data '{"config": {"url": "'${JENKINS_PROXY_URL}'/webhook/"}}' \
```
vpavlin: I think new url is not correct - should be `https://jenkins.prod-preview.openshift.io/github-webhook/`

vpavlin:
> How do we get this container to run in the cluster and inject the required properties?

My solution to this would be to add a template to deploy a job and map it to CM/Secret values existing in the cluster and manually deploy it

Very important feature request I'd have is a **dry run** option - i.e. instead of actual `curl -XPATCH` just print something like `updating $OLD_URL to $NEW_URL` - I think it would be worth it for final testing rounds

If not that, then I'd probably at least suggest adding ability to specify namespace(a) which you want to update - i.e. giving an env var `UPDATE_NAMESPACES=vpavlin hardy` and the script would respect that so we don't screw up everyone:)

WDYT?

scripts/webhook-migration/updateWebhooks.sh
hferentschik: That's what I  I thought. Leet me make a change then

scripts/webhook-migration/updateWebhooks.sh
hferentschik: I know :-), I wanted to get something working, but it is really not ideal. I'll use `jq`.

scripts/webhook-migration/updateWebhooks.sh
hferentschik: +1

hferentschik:
@vpavlin Ok, pushed an update addressing the discussed issues. Now would be time time to do a test run in the cluster.

internal/util/osio_token.go
```diff
@@ -0,0 +1,135 @@
...
+
+	token, err := generateToken(env, privateKey, user, uuid, session)
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("Unable to generate token for uuid '%s': %s", uuid, err))
+	}
+
+	return token, nil
+}
+
+func generateToken(env string, key *rsa.PrivateKey, user User, userID string, session string) (string, error) {
```
hferentschik: @alexeykazakov would you mind helping me with this token generation function? This function is a slightly modified version of your token-generation script, but I am having problems getting it to work on prod-preview. It spits out a token, but it does not seem to be valid when I then try to use it to get the GitHub token for this user.

internal/util/osio_token.go
```diff
@@ -0,0 +1,135 @@
...
+	token.Claims.(jwt.MapClaims)["name"] = user.Data.Attributes.FullName
+	token.Claims.(jwt.MapClaims)["company"] = ""
+	token.Claims.(jwt.MapClaims)["preferred_username"] = user.Data.Attributes.Username
+	token.Claims.(jwt.MapClaims)["given_name"] = ""
+	token.Claims.(jwt.MapClaims)["family_name"] = ""
+	token.Claims.(jwt.MapClaims)["email"] = user.Data.Attributes.Email
+
+	//	token.Claims.(jwt.MapClaims)["iat"] = time.Now().Unix()
+	token.Claims.(jwt.MapClaims)["allowed-origins"] = []string{
+		environments[env].osioUrl,
```
hferentschik: @alexeykazakov compared to your original code I changed these _allowed-origins_ URLs and adjusted them to match prod-preview. Other than that the method is pretty much unchanged. Does something else need to be adjusted or should I have kept the prod URLs?

internal/util/osio_token.go
hferentschik: FYI, the private key used is 'serviceaccount.privatekey'

hferentschik: Here is the error I get, when using the osio token as generated by this function:

    {"errors":[{"code":"unauthorized_error","detail":"There is no public key with such ID: 0lL0vXs9YRVqZMowyw8uNLR_yr0iFaozdQk9rzq2OVU","status":"401","title":"Unauthorized error"}]}

internal/util/osio_token.go
```diff
@@ -0,0 +1,136 @@
...
+		return "", errors.New(fmt.Sprintf("Unable to generate token for uuid '%s': %s", uuid, err))
+	}
+
+	return token, nil
+}
+
+func generateToken(env string, key *rsa.PrivateKey, user User, userID string, session string) (string, error) {
+	token := jwt.New(jwt.SigningMethodRS256)
+	//token.Header["kid"] = "0lL0vXs9YRVqZMowyw8uNLR_yr0iFaozdQk9rzq2OVU"
+	token.Header["kid"] = "PE6-BEECZZpPZIVxLR6NinbthOHJcGqYrfl8v7v6BVA"
```
alexeykazakov: Use the second kid from https://auth.prod-preview.openshift.io/api/token/keys for prod-preview
and the second kid from https://auth.openshift.io/api/token/keys for prod.

internal/util/osio_token.go
alexeykazakov: See my comment regarding kid (Key ID) below.

internal/util/osio_token.go
```diff
@@ -0,0 +1,136 @@
...
+
+	token.Claims.(jwt.MapClaims)["jti"] = uuid.NewV4().String()
+	token.Claims.(jwt.MapClaims)["exp"] = 1515627761
+	token.Claims.(jwt.MapClaims)["nbf"] = 0
+	token.Claims.(jwt.MapClaims)["iat"] = 1513035761
```
alexeykazakov: "iat" is "issued at"
So, use:
```
t := time.Now().Unix()
...
token.Claims.(jwt.MapClaims)["iat"] = t
token.Claims.(jwt.MapClaims)["auth_time"] = t
token.Claims.(jwt.MapClaims)["exp"] = t + 60*5 // in 5 minutes from now
```

hferentschik: got you thanks

hferentschik:
@vpavlin - Could you do one more review. I have been running in dry mode against the prod-preview cluster and I have been running against my account in live mode in prod_preview as well. 

Provided you think we are good, we should be able to merge. Then we sync with @gastaldi. His change to the launcher should go in first. @gastaldi where are you at with this change?

So the plan of action is:

1. Get the latest Proxy installed on prod-preview and prod
1. Get launcher installed with a patch for setting the new webhook URL
1. Run the migration scripts (part of this pull request)
1. Deploy a custom jenkins-sync-plugin - https://github.com/fabric8-services/fabric8-jenkins-proxy/issues/114

The latter is still work in progress.

gastaldi:
@hardy What change are you talking about specifically? I assume the one that uses the `JENKINS_URL` env var for every new booster deployed, right? If so that is already in production since yesterday ;)

hferentschik:
@gastaldi 

>  What change are you talking about specifically? I assume the one that uses the JENKINS_URL env var for every new booster deployed, right? If so that is already in production since yesterday ;)

Right. Good to know that it is in production. I take it prod-preview as well, right? Any chance you have a link to the issue and pull request in the launcher as a cross reference?

gastaldi:
>Right. Good to know that it is in production. I take it prod-preview as well, right? Any chance you have a link to the issue and pull request in the launcher as a cross reference?

Absolutely: The issue is https://github.com/fabric8-launcher/launcher-backend/pull/53 and the commit is https://github.com/fabric8-launcher/launcher-backend/commit/c897fae072566c729ff25cc25adde561b0690e2b

>And just to confirm, as of now, when a user gets provisioned and he sets up a repo via the wizard, he gets https://jenkins.prod-preview.openshift.io/github-webook resp. https://jenkins.openshift.io/github-webook as webhook target.

That's correct

scripts/webhook-migration/updateWebhooks.sh
```diff
@@ -0,0 +1,223 @@
...
+###############################################################################
+findGitHubRepos() {
+  export KUBECONFIG=/tmp/config
+  echo "Using KUBECONFIG ${KUBECONFIG}"
+  oc login ${OPENSHIFT_URL} --token=${OPENSHIFT_API_TOKEN} > /dev/null
+
+  local projects=$(oc projects -q)
+  for project in ${projects}; do
+   echo "Processing project ${project}"
+   oc project ${project} > /dev/null;
```
vpavlin: Not a big deal, but instead of switching projects, you could use `-n ${project}` in `oc` commands - it would speed the migration a bit, because (at least from my experience) switching the project takes quite a lot of time.

scripts/webhook-migration/updateWebhooks.sh
```diff
@@ -0,0 +1,223 @@
...
+   local tmp=($(oc get bc -o json | jq -r ' .items[] | select(.spec.source.git.uri) | .metadata.namespace + " " + .spec.source.git.uri[19:-4]'))
```
vpavlin: Also just realised - isn't `namespace == project` always?

scripts/webhook-migration/updateWebhooks.sh
```diff
@@ -0,0 +1,223 @@
...
+#   None
+###############################################################################
+getGitHubTokens() {
+  for i in "${!OSIO_TOKENS[@]}"; do
+    [ -z "${OSIO_TOKENS[${i}]}"  ] && echo "WARN: Empty OSIO token for namespace ${NAME_SPACES[$i]}" >&2 && continue
+
+    gh_token=$(curl -sgSL -H "Authorization: Bearer ${OSIO_TOKENS[${i}]}" ${AUTH_URL}api/token?for=https://github.com | jq -r '.access_token')
+    if [ "${gh_token}" == "null" ]; then
+        echo "WARN: Unable to create GitHub token for UUID ${UUIDS[$i]}" >&2
+        curl -sgSL -H "Authorization: Bearer ${OSIO_TOKENS[${i}]}" ${AUTH_URL}api/token?for=https://github.com >&2
```
vpavlin: Is this just to get output for the curl command to STDERR?

scripts/webhook-migration/updateWebhooks.sh
```diff
@@ -0,0 +1,223 @@
...
+# Checks whether the specified string ends with a '/' and if not adds one.
+# Globals:
+#   None
+# Arguments:
+#   $1       - The string to check for the trailing slash.
+# Returns:
+#   The unmodified string as passed to the function in case the string already has
+#   a trailing slash, otherwise the passed string with a '/' appended.
+###############################################################################
+ensureTrailingSlash() {
```
vpavlin: I'd probably do
```
ensureTrailingSlash() {
  echo $1 | sed 's#\([^/]\)#\1/#'
}
```
:)

vpavlin:
@hferentschik Added couple comments, but feel free to ignore them all - none of them actually changes how things work, just minor optimizations.

I'll try to run this tomorrow morning for my NS and repos and then we are probably good to go?

scripts/webhook-migration/updateWebhooks.sh
hferentschik: Ahh, good one. I was already wondering why this initial step was so slow. Let me try to use '-n'

scripts/webhook-migration/updateWebhooks.sh
hferentschik: Yes, I guess we could also use _namepace_ as the local variable name. Might be less confusing

scripts/webhook-migration/updateWebhooks.sh
hferentschik: Yes, I found this helpful for the cases where something failed.

scripts/webhook-migration/updateWebhooks.sh
hferentschik: LOL, not sure that is so much easier ;-) TBH, I usually reach for `sed` for these cases as well, but I found the current snippet when I did a search and it works and is more readable.

hferentschik:
> Added couple comments, but feel free to ignore them all - none of them actually changes how things work, just minor optimizations.

I definitely look into the way we iterate the namespaces. The time this takes was bothering me already (and that was just the prod-preview cluster). 

> I'll try to run this tomorrow morning for my NS and repos and then we are probably good to go?

+1

hferentschik:
I pushed an update. Now the script does not switch namespaces/projects all the time. It seems indeed faster. I also introduced the optional _CHECK_NAMESPACES_, so that one can run the script against a specific set of namespaces as well (for testing).

hferentschik:
If you want to run it against your namespace:

    oc process -p TARGET_ENV=stage -p DRY_RUN=true -p CHECK_NAMESPACES=vpavlin -f webhook-migration.job.yaml | oc apply -f -

vpavlin:
@hferentschik Had to add CHECK_NAMESPACES to the template, but otherwise looks good.

hferentschik:
Merged, thanks.

Issue #126, opened 2018-02-07
Request to Proxy timeouts before request to upstream


vpavlin:
With checks of route being available added because of #99, it sometimes happens that request to jenkins route does not finish before the request to proxy hits the timeout which then results in unexpected 504 to the loading page.

To fix this, we can use lower timeouts for the requests to Jenkins route

hferentschik:
What determines the timeout of a HTTP request in this case? Is this browser specific? There is no way of telling when a client request times out, right?

vpavlin:
I think it's combination of browser and OpenShift's HAProxy router, so not really.. My quick test showed we should get response from Jenkins in less than 1 s if it's up, so I'll set the timeout to 5 s, just to be sure there is enough head room

hferentschik:
Resolved via pull request #127

PR #127, opened 2018-02-07
Issue #130 Retry from loading page if status code is not 2xx

vpavlin:
As described in #126 , this PR handles the case where request Proxy -> Jenkins takes longer to return than User -> Proxy

I've also increased the time to re-check is Jenkins is still idled from 10 to 15 seconds - I think we could go even higher as we know unidling generally takes ~2-3 minutes

### DIFF
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"errors"
+
 	"github.com/fabric8-services/fabric8-jenkins-proxy/clients"
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/storage"
 	"github.com/patrickmn/go-cache"
@@ -415,7 +416,7 @@ func (p *Proxy) processTemplate(w http.ResponseWriter, ns string) (err error) {
 		Retry   int
 	}{
 		Message: "Jenkins has been idled. It is starting now, please wait...",
-		Retry:   10,
+		Retry:   15,
 	}
 	log.WithField("ns", ns).Debug("Templating index.html")
 	err = tmplt.Execute(w, data)
--- a/static/html/index.html
+++ b/static/html/index.html
@@ -6,6 +6,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>    
     <script type="application/javascript">
+      var timeout = {{.Retry}}*1000
       $(document).ready(function() {
         function check() {
           console.log("Checking "+window.location)
@@ -16,17 +17,21 @@
             statusCode: {
               202: function(){
                 console.log("Got 202, waiting along..")
-                setTimeout(check, {{.Retry}}*1000)
+                setTimeout(check, timeout)
                 },
               200: function() {
                 console.log("Got 200, reloading...")
                 window.location.reload(true)
                 }
+              default: function(xhr, status, e) {
+                console.log(status)
+                setTimeout(check, timeout)
+              }
             }          
           });
         }
 
-        setTimeout(check, {{.Retry}}*1000);
+        setTimeout(check, timeout);
       });
     </script>
     <style>


internal/proxy/proxy.go
```diff
@@ -28,6 +29,7 @@ const (
 	CookieJenkinsIdled = "JenkinsIdled"
 	ServiceName        = "jenkins"
 	SessionCookie      = "JSESSIONID"
+	RequestTimeout     = time.Second * 5
```
hferentschik: Let me see whether I understand this correctly. Client makes a request for the proxy with a certain timeout (I think Firefox uses something around 30 seconds or so). Our code will try to forward to the underlying Jenkins instance. To do this I makes a request to the Jenkins login page to make sure the Jenkins route is accessible (needed because the Jenkins pod might be running, but HAProxy has not updated the routes yet).

What happens is that the request to the login page actually does not return within the timeout limit of the original request, the original request times out with a 504 error.  

This pull request will change the timeout of the request to Jenkins to be quite low (5 seconds), so it will for sure timeout earlier than the original client request. But what does this mean for the user. He goes from error 504 to a error 500?

internal/proxy/proxy.go
```diff
@@ -307,8 +316,7 @@ func (p *Proxy) loginJenkins(pci ProxyCacheItem, osoToken string) (int, []*http.
 	} else {
 		log.WithField("ns", pci.NS).Infof("Accessing Jenkins route %s", jenkinsURL)
 	}
-	c := http.DefaultClient
-	resp, err := c.Do(req)
+	resp, err := p.httpClient.Do(req)
```
hferentschik: So when this request times out after 5 seconds, you get and err, right?

vpavlin: I think so

internal/proxy/proxy.go
vpavlin: Hmm..yeah, this PR fixes the "loading page" case, but potentially breaks the actual login case

hferentschik: So we need to think more about this again?

internal/proxy/proxy.go
hferentschik: Can we not test 'loginJenkins' in isolation and exercise the two main cases - once returning with success and once timing out? At least that would give some confidence in this method.

internal/proxy/proxy.go
vpavlin: thinking about it more - this  timeout change does not really change current behaviour - if there is problem with reaching Jenkins at the moment, there will be timeout anyway - in this case it only does not take 30s, but you get error after 5s.

There might be better way to handle things, I guess - add retries somewhere? Handle login error by redirect rather then returning error message?

hferentschik: > thinking about it more - this timeout change does not really change current behaviour - if there is problem with reaching Jenkins at the moment, there will be timeout anyway 

That was my thinking. You might change the error here 500 vs 504, but I am not sure. I have even not seen the 504 error yet, but admittingly I have not tried very hard either. I would like to get into a position where we can test the HTTP flow independent of the actual logic happening. On a very high coarse level, we could introduce a GitHubRequestHandler and a JenkinsRequestHandler (as interfaces). Proxy will be instantiated with implementations of these interfaces matching the current implementation. Proxy.Handle calls then just these interfaces depending on the type of request. Having this interface allows us to plug in dummy implementations of these handlers which don't call any external dependencies, but rather mock certain conditions which can occur. This would take some doing, but I don't see really any other way to test and fully understand what's going on.

> There might be better way to handle things, I guess - add retries somewhere? Handle login error by redirect rather then returning error message?

Probably, but I think we need to gain a better understanding first on what's going on. For now this seems a bit premature. I would for now probably close this pull request, but leave the issue itself open. As mentioned above I would start with some more code changes in order to get into a position where I can reproduce certain conditions.

vpavlin:
On the note of closing this - we need to put checks in place for the case, but it's not a pressing issue

Also from what I heard, OpenShift team is working hard on fixing the slow route addition issue, which will definitely help to solve this issue

vpavlin:
I removed the timeout change, but I'd like to keep the change to JavaScript - it will prevent the failure I saw as it would retry if it does not get 202 or 200 - unlike before when it just printed error in console and did not do anything

hferentschik:
@vpavlin Looking good. Could you, however, create a new issue for this and reference it in the commit message. This way we can keep issue #126 open.

vpavlin:
@hferentschik How do I retrigger the Travis build?

hferentschik:
Do we close this pull request then for now?

vpavlin:
I'd like to see this merged

hferentschik:
> I'd like to see this merged

Forget it. My wrong. I think I commented on the wrong pull request. It is merged.

PR #136, opened 2018-02-09
Push image build in from PR

vpavlin:


hferentschik:
Do we still want to merge this? What is the idea? Each pull request builds an image and pushes it? Then you can manually deploy it?

vpavlin:
Yeah, that was my idea to try out, but with your solution to run locally, we might not need to pollute the registry

hferentschik:
> Yeah, that was my idea to try out, but with your solution to run locally, we might not need to pollute the registry

I actually like this, but we can see how far we get with the ability to run locally. I created issue #147 to track this and we can always dig out this change of yours this way. Closing this pull request for now.

Issue #150, opened 2018-02-15
Proxy fails for users on 2a cluster


vpavlin:
Proxy now returns error for all tenants on our second OSO cluster

```
Got status 404 Not Found (404) from https://api.starter-us-east-2.openshift.com/oapi/v1/namespaces/jchevret-osiotest3-jenkins/deploymentconfigs/jenkins
```

This is due to the fact that Idler only knows about one cluster and thus asks wrong cluster for non-existing user

Temporary workaround is to return `false` in `IsIdle` call when a response `StatusCode` is `404`. In other words, if we cannot verify whether the Jenkins is idled, just carry on with actual proxying.

Side effect might be that a user who has never logged into OSIO and does not have any spaces (i.e. does not have OSO namespaces initilased) will get proxied to a non-existent route - which results in "Application Not Available" or some 5xx error - which is the same as if he would access the Jenkins route directly

hferentschik:
Sounds like a good plan. Let's go for that for now. Primarily we want to get Proxy/Idler rolled out for the single cluster case and then tackle the multi-cluster issue. There will be two alternatives we can look into.

hferentschik:
See also issue #151

vpavlin:
This will not work, because Proxy gets route from Idler by calling `GetRoute` (https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/master/internal/proxy/proxy.go#L387) and Idler will fail there as well

hferentschik:
Merged via pull request #154

PR #153, opened 2018-02-15
Issue #150 Making sure Proxy does not fail in multi-cluster scenario

hferentschik:
- Returning false for 404 reposnses from Idler.IsIdle
- Removing dependency to Idler for determining the Jenkins route
- Adding openshift.Client and all required wiring to use OpenShift API to query for route
- Moving clients into internal. There are no more external usages
- Removing obsolete tenant_test
- Updating setupLocalProxy.sh

Addresses issue #150


internal/openshift/openshift_client.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	return &OpenShift{
+		apiURL: apiURL,
+		token:  token,
+		client: client,
+	}
+}
+
+// GetRoute returns the OpenShift route for a given namespace and service and whether the route uses TLS.
+func (o *OpenShift) GetRoute(n string, s string) (string, bool, error) {
```
hferentschik: @vpavlin, on second thought. This might not be enoug.h. We moved this code from Idler to here, but the problem is still that GetRoute for a user in a different cluster won't resolve. There is more needed here. 

If we were to know the cluster URL of the namespace/user, we could adopt GetRoute to target the right cluster, but I am not sure how to do this. Effectively we will need to already partly implement some multi-cluster support :-(

hferentschik: As a super ugly workaround, we could for now manually construct the URL for the new cluster. AFAIU, this URL is pretty much static and it is just the namespace which differs. We could call GetRoute and if it does not return anything, we assume it is a user from the new cluster and construct the route accordingly.

I am even wondering do we really need in the long run to determine the route dynamically? It is more flexible in general, but would it be not enough to determine the target cluster and then build the URL by combining namespace and static cluster URL part?

hferentschik: Or we somehow use this OSO Proxy, but I am not sure what it does and how to use it.

vpavlin: Right, this is not enough, we need to be able pass OSO api url to GetRoute call.

We can determine cluster url from GetTenantInfo call which is used in both flows (webhook and UI) - we already store the cluster url in ProxyCacheItem structure

hferentschik: ahh, cool. Seems you have a plan :-) What's about the token? Is it the same for all clusters? Is it enough to just pass the OSO API URL and reuse the token? Thinking back to our conversation yesterday, I think that's what you always said, but I did not quite understand ;-)

hferentschik: So the interface would become GetRoute(osoURL string, namespace string, service string), right? Assuming the token is the same and we do already know which cluster a tenant is provisioned on, it should indeed be easy. I guess we also remove _apiURL_ from the OpenShift struct.

Sounds like a plan. I might be able to work today, but we can sync later. If you want to get started, go ahead. I also need to do several things first. I will check-in with you first before doing anything more on this.

hferentschik:
Closing this pull request in favour of #154

PR #154, opened 2018-02-16
Issue #150 Making sure Proxy does not fail in multi-cluster scenario

vpavlin:
We need to get list of clusters to generate routes to Jenkins. We will read them from fabric8-auth secrets - we only load apiURL and appDNS keys from the secret and then use these to generate Jenkins route URL based on the cluster user belongs to


internal/configuration/configuration.go
```diff
@@ -32,20 +35,39 @@ const (
...
 	varMaxRequestRetry = "max.request.retry"
 
+	varOpenShiftApiURL   = "openshift.api.url"
+	varOpenShiftApiToken = "openshift.api.token"
+
 	varIndexPath = "index.path"
 
-	varLocalDevEnv = "local.dev.env"
+	osoClusterConfigFileName    = "oso-clusters.conf"
+	defaultOsoClusterConfigPath = "/etc/fabric8/" + osoClusterConfigFileName
```
vpavlin: Should we make this path configurable? (if nothing else, it would simplify testing, so I assume answer is yes:) )

hferentschik: Seems this also causes test failures

internal/clients/tenant.go
```diff
@@ -108,11 +109,11 @@ func (t Tenant) GetTenantInfo(tenantID string) (ti TenantInfo, err error) {
 }
 
 // GetNamespaceByType searches tenant namespaces for a given type
-func (t Tenant) GetNamespaceByType(ti TenantInfo, typ string) (r *Namespace, err error) {
+func (t Tenant) GetNamespaceByType(ti TenantInfo, typ string) (r Namespace, err error) {
```
hferentschik: +1

internal/configuration/configuration.go
```diff
@@ -32,20 +35,39 @@ const (
 	varKeycloakURL     = "keycloak.url"
 	varMaxRequestRetry = "max.request.retry"
 
+	varOpenShiftApiURL   = "openshift.api.url"
```
hferentschik: These can be removed now. They are unused

internal/configuration/configuration.go
hferentschik: So we most likely need to inject the location as env variable

internal/configuration/configuration.go
```diff
@@ -146,6 +168,30 @@ func (c *Data) GetPostgresConfigString() string {
 	)
 }
 
+func loadOSOClusters() (map[string]string, error) {
+	if _, err := os.Stat(defaultOsoClusterConfigPath); err != nil {
```
hferentschik: ok, let's do this then from reading an ENV variable. Using the mounted secret seems a bit "dangerous" and a bit of an overkill.

internal/proxy/proxy.go
```diff
@@ -70,6 +72,7 @@ type Proxy struct {
 	storageService  storage.Store
 	indexPath       string
 	maxRequestRetry int
+	clusters        map[string]string
```
hferentschik: +1

internal/proxy/proxy.go
```diff
@@ -620,6 +618,19 @@ func (p *Proxy) RecordStatistics(ns string, la int64, lbf int64) (err error) {
...
 
+func (p *Proxy) getScheme(tlsEnabled bool) string {
+	if tlsEnabled {
+		return "https"
+	}
+	return "http"
+
+}
+
+func (p *Proxy) constructRoute(clusterURL string, ns string) (string, string) {
```
hferentschik: +1

internal/proxy/proxy.go
```diff
@@ -620,6 +618,19 @@ func (p *Proxy) RecordStatistics(ns string, la int64, lbf int64) (err error) {
 	return
 }
 
+func (p *Proxy) getScheme(tlsEnabled bool) string {
```
hferentschik: I think this can be removed

openshift/jenkins-proxy.app.yaml
```diff
@@ -29,6 +29,16 @@ objects:
       spec:
         containers:
         - env:
+          - name: JC_OPENSHIFT_API_URL
```
hferentschik: This can also be removed

openshift/jenkins-proxy.app.yaml
```diff
@@ -29,6 +29,16 @@ objects:
...
+            valueFrom:
+              configMapKeyRef:
+                name: f8tenant
+                key: openshift.tenant.masterurl
+          - name: JC_OPENSHIFT_API_TOKEN
```
hferentschik: And this

openshift/jenkins-proxy.app.yaml
```diff
@@ -126,6 +136,17 @@ objects:
             limits:
               memory: "128Mi"
               cpu: "500m"
+          volumeMounts:
```
hferentschik: Ok, let's replace this with an injected variable for now.

scripts/setupLocalProxy.sh
```diff
@@ -122,12 +136,26 @@ start() {
...
+# Arguments:
+#   None
+# Returns:
+#   None
+###############################################################################
+env() {
+    [ -z "${JC_OPENSHIFT_API_TOKEN}" ] && printHelp && exit 1
+    [ -z "${JC_AUTH_TOKEN}" ] && printHelp && exit 1
 
+    echo export JC_OPENSHIFT_API_URL=https://api.free-stg.openshift.com
```
hferentschik: we probably should export here the variable for the cluster config

internal/proxy/proxy.go
```diff
@@ -620,6 +625,16 @@ func (p *Proxy) RecordStatistics(ns string, la int64, lbf int64) (err error) {
 	return
 }
 
+//constructRoute returns Jenkins route based on a specific pattern
+func (p *Proxy) constructRoute(clusterURL string, ns string) (string, string, error) {
+	appSuffix := p.clusters[clusterURL]
+	if len(appSuffix) == 0 {
```
vpavlin: @hferentschik Added check for missing cluster

hferentschik:
Merged, thanks.

Issue #159, opened 2018-02-16
Proxy client needs to query correct namespace


vpavlin:
With last changes Proxy queries Idler for tenant name rather than Jenkins namespace (`$tenant-jenkins`)

hferentschik:
Merged via pull request #160

PR #160, opened 2018-02-16
Issue #159 Idler client needs to query correct namespace

vpavlin:
Add `-jenkins` suffix to namespace if needed

### DIFF
--- a/internal/clients/idler.go
+++ b/internal/clients/idler.go
@@ -5,8 +5,13 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
+)
+
+const (
+	namespaceSuffix = "-jenkins"
 )
 
 //Idler is a simple client for Idler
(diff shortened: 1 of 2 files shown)


internal/clients/idler.go
```diff
@@ -24,8 +29,13 @@ type Status struct {
 	IsIdle bool `json:"is_idle"`
 }
 
-//IsIdle returns true if Jenkins is idled for a given namespace
-func (i Idler) IsIdle(namespace string) (bool, error) {
+// IsIdle returns true if the Jenkins instance for the specified tenant is idled. False otherwise
```
hferentschik: +1

hferentschik:
Merged.

internal/clients/idler.go
```diff
@@ -24,8 +29,13 @@ type Status struct {
...
+func (i Idler) IsIdle(tenant string) (bool, error) {
+	namespace := tenant
+	if !strings.HasSuffix(tenant, namespaceSuffix) {
+		namespace = tenant + namespaceSuffix
```
aslakknutsen: This should really rely on the Tenant Service provided namespace name of type "jenkins" instead of hardcoding a naming pattern

hferentschik: Can you create an issue and also point to the API to use?

Issue #161, opened 2018-02-16
Proxy needs to explicitly call unidle 


hferentschik:

hferentschik:
Merged via pull request #164

PR #164, opened 2018-02-16
Issue #161 Calling explicitly unidle when Jenkins in idle state

hferentschik:
Fix for issue #161

### DIFF
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -223,6 +223,12 @@ func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request, r
 
 		//Break the process if the Jenkins is idled, set a cookie and redirect to self
 		if isIdle {
+			err = p.idler.UnIdle(ns)
+			if err != nil {
+				p.HandleError(w, err, requestLogEntry)
+				return
+			}
+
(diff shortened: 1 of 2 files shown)


internal/proxy/proxy.go
```diff
@@ -223,6 +223,12 @@ func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request, r
 
 		//Break the process if the Jenkins is idled, set a cookie and redirect to self
 		if isIdle {
+			err = p.idler.UnIdle(ns)
```
hferentschik: I guess this is correct. If we cannot make the unidle call, no need to set the cookie, right?

internal/proxy/proxy.go
```diff
@@ -408,6 +419,10 @@ func (p *Proxy) handleGitHubRequest(w http.ResponseWriter, r *http.Request, requ
 	//If Jenkins is idle, we need to cache the request and return success
 	if isIdle {
 		p.storeGHRequest(w, r, ns, body, requestLogEntry)
+		err = p.idler.UnIdle(ns)
```
hferentschik: here i actually store the request first and then try to unidle. WDYT?

Issue #172, opened 2018-02-21
Use a session state to generate OSIO tokens for accounts which failed initial webhook migration


hferentschik:

hferentschik:
Merged via pull request #177

PR #173, opened 2018-02-22
Issue #172 Consume SESSION from environment

vpavlin:
We don't know if it fixes the issue yet


hferentschik:
We will have to change the private key id as well, right?

hferentschik:
We also need to take care of the private key id. I made some additional changes to cater for that in pull request #177.

Closing this pull request in favour of #177

Issue #122, opened 2018-01-26
Change papi to api in the API endpoint URL


hferentschik:
The use of _papi_ is a relict of some old code. Really it should be _api_. The change is trivial, but it requires a change on the Idler at the same time and then a synced re-deploy.

hferentschik:
Merged via pull request #183

PR #183, opened 2018-03-06
Issue #122 let change the `papi` URL path to `api`

chmouel:
As documented in #122 let change the `papi` URL path to `api`

### DIFF
--- a/internal/router/router_test.go
+++ b/internal/router/router_test.go
@@ -1,0 +1,66 @@
+package router
+
+import (
+	"bytes"
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/storage"
+	"github.com/julienschmidt/httprouter"
+	"github.com/stretchr/testify/assert"
+)
+
+type mockResponseWriter struct {
+	buffer bytes.Buffer
+}
+
+func (m *mockResponseWriter) Header() (h http.Header) {
+	return http.Header{}
+}
+
+func (m *mockResponseWriter) Write(p []byte) (n int, err error) {
+	m.buffer.Write(p)
+	return len(p), nil
+}
+
+func (m *mockResponseWriter) WriteString(s string) (n int, err error) {
+	m.buffer.WriteString(s)
+	return len(s), nil
+}
+
+func (m *mockResponseWriter) WriteHeader(int) {}
+
+func (m *mockResponseWriter) GetBody() string {
+	return m.buffer.String()
+}
+
+type mockProxyAPI struct {
+	storageService storage.Store
+}
+
+func (i *mockProxyAPI) Info(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+	w.Write([]byte("Info"))
+	w.WriteHeader(http.StatusOK)
+}
+
+func Test_all_routes_are_setup(t *testing.T) {
+	mockedProxyApi := &mockProxyAPI{}
+	mockedRouter := CreateAPIRouter(mockedProxyApi)
+
+	var routes = []struct {
+		route  string
+		target string
+	}{
+		{"/api/info/:namespace", "Info"},
+	}
+
+	for _, testRoute := range routes {
+		w := new(mockResponseWriter)
+
+		req, _ := http.NewRequest("GET", testRoute.route, nil)
+		mockedRouter.ServeHTTP(w, req)
+
+		assert.Equal(t, testRoute.target, w.GetBody(), fmt.Sprintf("Routing failed for %s", testRoute.route))
+	}
+}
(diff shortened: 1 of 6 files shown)


chmouel:
Useful addition for newcommers to add to the git hooks : 

```
$ cat .git/hooks/pre-push                                                                                         
#!/bin/bash

set -e

make validate_commits
chmouel@TheBatman:github.com/fabric8-services/fabric8-jenkins-proxy$
```

hferentschik:
Just run 'make all'. If 'make all' does not pass locally, you probably don't have a pull request yet ready to be merged. The problem is that in some cases one wants to push a WIP in order to get some feedback. In this case your commit hook can get into the way.

hferentschik:
I am missing some tests here. Similar to the Idler side we should write a unit test which prevents us to accidentally change routes https://github.com/fabric8-services/fabric8-jenkins-idler/blob/faff35d02625cc97b542c76c988742a0a7860a5e/internal/router/router_test.go#L72

Also, this issue needs to be addressed together with the Idler. See also https://github.com/fabric8-services/fabric8-jenkins-idler/issues/145. Merging this w/o deploying an updated Idler will break things. Remember, a merge to master deploys to prod-preview. You merge this and you effectively break prod-preview. 

This pull request should also change https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/7330d9ee78e9ce84e4f13b6efb5562286f817834/internal/clients/idler.go#L40 and then as part of ttps://github.com/fabric8-services/fabric8-jenkins-idler/issues/145 are are going to change the Idlers API as well as the API of the Proxy which the Idler is calling.

hferentschik:
add to whitelist

chmouel:
Added an issue for the tests here https://github.com/fabric8-services/fabric8-jenkins-proxy/issues/184

hferentschik:
> Added an issue for the tests here #184

Why would you create a new issue for that? The whole point is that one develops tests as part of am bug fix / new feature. The test should be part of this pull request. It is directly related. 

Usually a pull request contains a fix/feature + tests + and required doc changes.

I am going to close #184

hferentschik:
I am confused. You added the tests after all? So I am still going to close #184 and maybe you can just roll this pull request into a single commit for issue #122

chmouel:
The tests are additions, this PR works fine as is if only the tests add more robustness so then is the PR but to everyone their own.

chmouel:
So you want to change /iapi to /api too ? this isn't clear from the title of this issue

hferentschik:
I might have not been clear. I thought I made a comment about it, but it might have been lost or misunderstood. Also if one reads the title of https://github.com/fabric8-services/fabric8-jenkins-idler/issues/145 more closely it also gives a hint. But agreed, I should have been more explicit.

hferentschik:
> The tests are additions, this PR works fine as is 

The tests give additional confidence in that the issue/bug got addressed. IMO they belong to the fix and not a different issues. The aim should always be to provide a test for each change. Obviously this cannot be always achieved. AFAIU @bartoszmajsak for example is working on a pull request integration which will actually deny the merge of a pull request, unless it contains a test. Only if a maintainer explicitly accepts a pull request without tests a merge can be done.

chmouel:
sure,  I have just added those tests, lmk if there is anything else to do for those PR

internal/router/router_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/storage"
+	"github.com/julienschmidt/httprouter"
+	"github.com/stretchr/testify/assert"
+)
+
+// TODO(chmouel): Implement a memory store which could be used between all tests instead of having to mock it everywhere
```
hferentschik: I agree, but why don't you create a GitHub issue to track this, instead of a TODO in the code?

internal/router/router_test.go
```diff
@@ -0,0 +1,113 @@
...
+type mockStore struct {
```
hferentschik: There is no need for mocking the store here. You are not making any use of it.

kishansagathiya:
@hferentschik Any specific reason for having some methods in camelCase and some with underscore https://github.com/fabric8-services/fabric8-jenkins-idler/blob/faff35d02625cc97b542c76c988742a0a7860a5e/internal/router/router_test.go#L72?

internal/router/router_test.go
chmouel: 👍 thanks

chmouel:
@kishansagathiya imho tests names are usually long strings, underscores would make sense here, go functions that are not tests should be camelcase (as advised [here](https://golang.org/doc/effective_go.html#mixed-caps))

hferentschik:
> Any specific reason for having some methods in camelCase and some with underscore

Yes, it is a style thing. I usually use underscores, but do not enforce it. I find things like test function names like 'TestMethodX' pointless. They often tell me not much. I rather have longer function names which describe in prose what the test is about. In this case when the function names get longer and more prose like '_' makes it more readable. If you run tests with '-v' enabled and look at the output you can already "read" what's getting tested. This is a lot based on K. Henney's tips around writing good tests.

hferentschik:
The code looks good now.  I am going to approve the pull request, but keep the WIP status until https://github.com/fabric8-services/fabric8-jenkins-idler/pull/146 is in place.

hferentschik:
Merged, thanks

Issue #192, opened 2018-03-12
Proxy should retrieve cluster view from Idler


hferentschik:
Instead of the statically injected `JC_OSO_CLUSTERS`, the Proxy should call the newly created Idler API `ClusterDNSView`. This needs to happen during startup and in case of failure, the startup should fail.

`JC_OSO_CLUSTERS` can then be removed.

hferentschik:
Merged via pull request #194

Issue #187, opened 2018-03-12
Calls to the Ider API should send the cluster URL as additional paramters


hferentschik:
The affected code lives in `idler.go` where the Proxy is calling the Idler API `Isdle` and `UnIdle`. As part of these call the additional parameter `openshift_api_url` should be passed along. 

Looking briefly at the code, I believe we have the information already, we just don't use it.

hferentschik:
Merged via pull request #194

PR #194, opened 2018-03-13
Issue #187 Making sure all calls to Idler API pass OpenShift API URL

hferentschik:
Addresses issue #187 and #192 
    
- Issue #192 Making sure Proxy retrieves cluster view from Idler
  - Removing obsolete config options
  - Updating README and adding info about how to trigger webhooks
  - Some logging improvements

- Issue #187 Making sure all calls to Idler API pass OpenShift API URL


hferentschik:
@vpavlin , do you mind another sanity check :-) This is basically the Proxy side of multi-cluster. Everything was pretty much straightforward. In `ProcessBuffer` I had to jump through a few more hoops to get back the OpenShift API URL. It would have been easier if we had it in the DB as well, but it is not. I wanted to avoid adding all the pieces to do managed database migrations, so I ended up making the required calls to determine the missing URL. Seems to work fine.

hferentschik:
Merged

Issue #174, opened 2018-02-22
Local Proxy testing is impossible due to insecure protocol


vpavlin:
When you run proxy locally with `setupLocalProxy.sh` it uses http which means cookies coming from Jenkins (which are used to verify user is logged in) will not be stored and will be thrown away by browser as they require `secure` connection (https://tools.ietf.org/html/rfc6265#section-4.1.2.5):

```
JSESSIONID.fa4b186c=node01o0uvpmvnknal1sja4z9lhvy8w323.node0; Path=/; HttpOnly; Secure
```

That means that for the local testing we need to somehow generate TLS certs and start with https - it was originally there in [testrun.go](https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/c705bb1b3b3a6b544deda9adc90e696cdfeabeea/testutils/testrun.go#L78), but it required the certs to be already present, it would be great if we could add a function to `setupLocalProxy.sh` to generate them

hferentschik:
> That means that for the local testing we need to somehow generate TLS certs and start with https 

What exactly needs to start with HTTPS? The Proxy itself?

vpavlin:
Yes

kishansagathiya:
closed with https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/264

PR #198, opened 2018-03-16
Issue #179 Refactor and test handleJenkinsUIRequest()

### LINKED ISSUE #179: Refactor Proxy.handleJenkinsUIRequest
* Split method into smaller parts
* Untangle the mixture of returning values (_needsAuth_, _noProxy_) as well as writing to the response writer on the fly
* Add unit tests

kishansagathiya:
Fixes #179


hferentschik:
So that for now is just a variable rename?

kishansagathiya:
@hferentschik 
No, 
The code from handleJenkinsUIRequest() has been transferred to processTokenJSON() and checkCookies().
Does this refactor seem right to you?
If this is the kind of refactor that you were expecting I can move to writing tests.

internal/proxy/proxy.go
```diff
@@ -192,165 +192,189 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
...
 	//If the user provides OSO token, we can directly proxy
-	if _, ok := r.Header["Authorization"]; ok { //FIXME Do we need this?
+	if _, ok := request.Header["Authorization"]; ok { //FIXME Do we need this?
 		needsAuth = false
 	}
 
-	if tj, ok := r.URL.Query()["token_json"]; ok { //If there is token_json in query, process it, find user info and login to Jenkins
-		if len(tj) < 1 {
-			p.HandleError(w, errors.New("could not read JWT token from URL"), requestLogEntry)
+	namespace, noProxy = p.processTokenJSON(responseWriter, request, requestLogEntry, redirectURL)
```
hferentschik: Ok, now I get it. It is a start, but I don't think there is a need for returning anything here. Especially look at  p.checkCookies. 

I guess as a first step this works, but the next is to eliminate the returns. Thale the last if statement for example `if needsAuth`. This could be pulled out into a function which then can get called.

Also, we tried to remove most of the used naked returns in the code. In most places we have done it already, but the code in `proxy.go` uses still a lot of them. The use of naked returns is really discouraged and should if ever be only used for very small methods. IMO in the current code it just makes it hard to understand. Let's get rid of them.

hferentschik: That said, I think we need to keep some conditional there. Pulling the conditionals into the methods obfuscates the code. Now it looks like, you are always processing the JSON token and you are always proecssing the cookies. There are several phases here which are not obvious anymore. 

Leaving the outer if statements here will makes things clearer

internal/proxy/proxy.go
```diff
@@ -192,165 +192,189 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
...
+		http.Redirect(responseWriter, request, redirAuth, 301)
+	}
+	return
+}
+
+// processTokenJSON processes token_json if present in query and find user info and login to Jenkins
+func (p *Proxy) processTokenJSON(responseWriter http.ResponseWriter, request *http.Request, requestLogEntry *log.Entry, redirectURL *url.URL) (namespace string, noProxy bool) {
+
+	// Checks if token_json is present in the query
+	if tokenJSON, ok := request.URL.Query()["token_json"]; ok {
```
hferentschik: It might make sense to extract this into something like `func isAuthenticatedRequest(request *http.Request) bool`. If you then pull up the if statement again, the method name documents the alogrithm.

```golang
func (p *Proxy) handleJenkinsUIRequest(responseWriter http.ResponseWriter, request *http.Request, requestLogEntry *log.Entry) (cacheKey string, namespace string, noProxy bool) {
        if isAuthenticatedRequest(request) {
	   namespace, noProxy = p.processAuthenticatedRequest(responseWriter, request, requestLogEntry, redirectURL)
        } else {
           return redirectToAuth(...)
       }
 
        if isWaitLoopRequest(request) {
	    cacheKey, namespace, noProxy, needsAuth = p.checkCookies(responseWriter, request, requestLogEntry)
         }
}
```

internal/proxy/proxy.go
```diff
@@ -192,165 +192,189 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
...
 
-func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request, requestLogEntry *log.Entry) (cacheKey string, ns string, noProxy bool) {
+func (p *Proxy) handleJenkinsUIRequest(responseWriter http.ResponseWriter, request *http.Request, requestLogEntry *log.Entry) (cacheKey string, namespace string, noProxy bool) {
 	//redirect determines if we need to redirect to auth service
 	needsAuth := true
 	noProxy = true
 
-	//redirectURL is used for auth service as a target of successful auth redirect
-	redirectURL, err := url.ParseRequestURI(fmt.Sprintf("%s%s", strings.TrimRight(p.redirect, "/"), r.URL.Path))
+	// redirectURL is used for auth service as a target of successful auth redirect
```
hferentschik: I don't think we need to do that here. We can do that where it is needed. And most likely it can also happen in a dedicated properly named function

internal/proxy/proxy.go
```diff
@@ -192,165 +192,189 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
...
+	redirectURL, err := url.ParseRequestURI(fmt.Sprintf("%s%s", strings.TrimRight(p.redirect, "/"), request.URL.Path))
 	if err != nil {
-		p.HandleError(w, err, requestLogEntry)
+		p.HandleError(responseWriter, err, requestLogEntry)
 		return
 	}
 
 	//If the user provides OSO token, we can directly proxy
-	if _, ok := r.Header["Authorization"]; ok { //FIXME Do we need this?
+	if _, ok := request.Header["Authorization"]; ok { //FIXME Do we need this?
```
hferentschik: For the record, this can be removed. Needs to be tested though.

hferentschik:
While working on this you might want to verify by running the Proxy locally whether it still works. There might be an issue with cookies in this case. I have not confirmed that and I successfully did a proxying using a local Proxy instance, but there is this issue which might prevent you to test the workflow locally https://github.com/fabric8-services/fabric8-jenkins-proxy/issues/174. Might be worth following up. 

Being able to test the workflow locally will help a lot, but remember the end game is to "codify" the flow into some tests.

hferentschik:
Any more changes to this?

internal/proxy/proxy.go
kishansagathiya: Not sure which code segment could be kept in isWaitLoopRequest(request)

kishansagathiya:
Is this(https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/master/internal/proxy/proxy.go#L315) the function that waits until jenkins is unidled?

kishansagathiya:
@chmouel If this seems right to you I can move forward with tests. Except for breaking the function this doesnt have much changed.

chmouel:
@kishansagathiya there is a few concernces about this from @hferentschik what are about?

internal/proxy/proxy.go
kishansagathiya: https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/198/commits/b4bc94048cc26b388ce3358165760aff2459154f

internal/proxy/proxy.go
kishansagathiya: https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/198/commits/b4bc94048cc26b388ce3358165760aff2459154f

kishansagathiya:
@chmouel I think they are all covered
- No more naked returns
- No more Authorization Header
- redirectURL is in smaller functions
- less returned values
- conditions are out of the methods

kishansagathiya:
Blocked with https://github.com/fabric8-services/fabric8-jenkins-proxy/issues/203

kishansagathiya:
@chmouel This is NOT ready for merge yet. It is WIP. We can merge this after tests. The reason for such temporary commits is to make sure that I am on the right path. If I get a review after tests, I will have change implementation and tests both.

kishansagathiya:
@chmouel Have added some tests. It needs much more, which I keep pushing with this PR. 
btw do I need to mock the database as well? not sure, what should I do? fabric8-auth doesn't mock db.

chmouel:
look at the way it uses https://github.com/ory/dockertest for the database tests, there is already some example in the in db_store_test [here](https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/988d21c01569b7276949763ac197aab1a936c93f/internal/storage/db_store_test.go#L24)

kishansagathiya:
Tested this locally to be working using
```
http://localhost:8080/api/info/ksagathi-preview
```
It unidled the Jenkins and this is on running it again.

```
{"component":"proxy","header":"-H Accept-Encoding: gzip, deflate, br -H Accept-Language: en-US,en;q=0.9 -H Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8 -H Cache-Control: max-age=0 -H Connection: keep-alive -H Cookie: JenkinsIdled=1d59256b-1e01-4314-8580-054849972197 -H Upgrade-Insecure-Requests: 1 -H User-Agent: Mozilla/5.0 (X11; Fedora; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/63.0.3239.132 Safari/537.36","level":"info","msg":"Handling incoming proxy request.","request":"GET /api/info/ksagathi-preview","request-hash":<hidden>,"time":"2018-04-03T16:37:41+05:30","type":"Jenkins UI"}
{"component":"proxy","level":"info","msg":"Accessing Jenkins route https://jenkins-ksagathi-preview-jenkins.b542.starter-us-east-2a.openshiftapps.com/securityRealm/commenceLogin?from=%2F","ns":"ksagathi-preview-jenkins","request-hash":<hidden>,"time":"2018-04-03T16:37:41+05:30"}
{"component":"proxy","level":"info","msg":"Found cookie <hidden>","ns":"ksagathi-preview-jenkins","request-hash":<hidden>,"time":"2018-04-03T16:37:42+05:30"}

```

chmouel:
@kishansagathiya fair enough, feel free to address the mock package move and i'll be 👍 with this,

kishansagathiya:
@chmouel on it

kishansagathiya:
@chmouel 
Moving mocks to other package was very hairy

- It creates import cycle
- Removing import cycle would require us to move tests to a separate package say something like proxy_test, which would require us to export a lot of structs, interfaces, and methods
- I requires to move other tests also like move db_store_test to separate package. I have those changes, but thought just moving mocks to different file would be a much better solution.

PR #204, opened 2018-03-23
Issue #133 NotFound should be an error

### LINKED ISSUE #133: Error in ProxyAPI.Info
When turning on debug trace, we get a whole bunch of:

    {"level":"debug","msg":"Did not find data for none","time":"2018-02-09T11:00:21Z"}

Looking at the underlying code I am also wondering a bit about the implementation:

```Go
func (api *ProxyAPI) Info(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
	ns := ps.ByName("namespace")
	s, notFound, err := api.storageService.GetStatisticsUser(ns)
	if err != nil {
		if notFound {
			log.Debugf("Did not find data for %s", ns)
		} else {
			log.Error(err) //FIXME
			fmt.Fprintf(w, "{'error': '%s'}", err)
			return
		}
	}
	c, err := api.storageService.GetRequestsCount(ns)
	if err != nil {
		log.Error(err) //FIXME
		fmt.Fprintf(w, "{'error': '%s'}", err)
		return
	}
	resp := APIResponse{
		Namespace:   ns,
		Requests:    c,
		LastRequest: s.LastBufferedRequest,
		LastVisit:   s.LastAccessed,
	}

	json.NewEncoder(w).Encode(resp)
}
```

If there is no record found, will there be also an error? That seems odd.

kishansagathiya:
Now if namespace is not found info api will through an error

Fixes #133

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -37,15 +37,12 @@ type Response struct {
 // Info returns JSON including information about Proxy usage statistics for a given namespace.
 func (api *proxy) Info(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	ns := ps.ByName("namespace")
-	s, notFound, err := api.storageService.GetStatisticsUser(ns)
+	s, _, err := api.storageService.GetStatisticsUser(ns)
 	if err != nil {
-		if notFound {
-			log.Debugf("Did not find data for %s", ns)
-		} else {
-			log.Error(err) //FIXME
-			fmt.Fprintf(w, "{'error': '%s'}", err)
-			return
-		}
+		log.Error(err)
+		fmt.Fprintf(w, "{'error': '%s'}", err)
+		return
+
 	}
 	c, err := api.storageService.GetRequestsCount(ns)
 	if err != nil {


kishansagathiya:
@chmouel WDYT?

chmouel:
Can you please word a bit more the issue and want you try to achieve in the commit, i don't quite understand it,

kishansagathiya:
@chmouel  Should we throw an error if namespace passed with info api is not found? So, far we weren't. This PR changes that.

this would happen when we run GET /api/info/$SOME_NAMESPACE_THAT DOESNT_EXIST

chmouel:
Sure but please update the title of this commit/issue then and description,

kishansagathiya:
Let's keep this one on HOLD.  There are places where if notFound, it records new statistics considering it as a new user

kishansagathiya:
Ok. Turns out notFound should not throw an error. NotFound means it is not found in the database. If the namespace is not in the database, it will add the namespace to the database. It is not supposed to be an error. Let's close this PR and the issue as well. @chmouel WDYT?

Some uses of notFound
https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/8338ca57095f76d445dbfc6cb02fbe44052dbf84/internal/api/api.go#L40
https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/8338ca57095f76d445dbfc6cb02fbe44052dbf84/internal/proxy/proxy.go#L615

chmouel:
You know better this code than me :) if you tell me this can be closed, so be it, but yeah from your explanation this should not be indeed an error.

kishansagathiya:
If you look at this you will find that if notFound, it creates newStatistics

https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/8338ca57095f76d445dbfc6cb02fbe44052dbf84/internal/proxy/proxy.go#L615-L633

Issue #231, opened 2018-04-13
Create Prometheus metrics for accessing counter


lordofthejars:
Create Prometheus metrics for how many times access to Jenkins UI are accessed and how many times of those access are coming from github webhooks.

lordofthejars:
https://github.com/openshiftio/openshift.io/issues/2778

kishansagathiya:
@lordofthejars 
Done with https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/232, right?

PR #232, opened 2018-04-13
Issue #231 Add Prometheus metrics for accessing type counter

lordofthejars:


lordofthejars:
@nurali-techie If you can review it :)

chmouel:
I don't know much about prometheus but i trust you on this has been tested and working, 👍 after the rebase,

internal/proxy/proxy.go
```diff
@@ -134,6 +141,8 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
 		requestType = "Jenkins UI"
 	}
 
+	Recorder.RecordReqByTypeTotal(requestType)
```
nurali-techie: It would be good if you can move metric recording to some middlewares.  Similar (not exactly) to https://github.com/fabric8-services/fabric8-wit/blob/master/metric/recorder.go#L20.  So logic of metrics recording will be separated from actual request handling logic which consider clean design and it has it's benefits.

internal/proxy/proxy.go
```diff
@@ -134,6 +141,8 @@ func (p *Proxy) Handle(w http.ResponseWriter, r *http.Request) {
 		requestType = "Jenkins UI"
```
nurali-techie: It's better to have metrics lable's value without space.  Also, I have seen to keep value in lower case always.

nurali-techie: Also, I understand that you want to count how many request comes from GitHub Web-Hook and from OSIO .. right .. ?
If so, why there is request type called 'Jenkins UI'.  Or it is NOT coming from OSIO?

internal/proxy/proxy.go
nurali-techie: Also, I see that prometheus is registered at - https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/master/internal/router/router.go#L14
Actually, prometheus.Handler() is deprecated.  If possible, use promhttp.Handler().  Check - https://github.com/fabric8-services/fabric8-wit/blob/master/main.go#L416.

nurali-techie:
sorry @lordofthejars for reviewing your PR late.  Actually I was involved in multiple prometheus PR and I overlook yours.
I have provided few review comments.  Functionally, your code looks good :-)  There are few design things which you can do.  Ping me anytime if you want more details on how to in-corporate my design review comments.

internal/proxy/proxy.go
lordofthejars: Well in code itself is identified as coming from Jenkins or coming from WebHook, so I decided to use same names.

lordofthejars: Well notice that these labels were already there, for jenkins-proxy a request is either a webhook or Jenkins UI so I decided to follow same convention so in metrics and in proxy itself maintains same name. But as you said I am going to put everything in lowercase and without space.

Issue #241, opened 2018-04-25
Update Prometheus code


lordofthejars:
Update Prometheus code with next comments: https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/232#pullrequestreview-115098624

PR #242, opened 2018-04-25
Issue #241 Refactor metrics code for having better label names

lordofthejars:


lordofthejars:
@nurali-techie I have updated the code with some of your suggestions. Thank you very much.

Notice that I have not added the code into a middleware thingy. The reason I choose to not do it is because we are publishing one value that it is calculated (if comes from GH or Jenkins) and notice that this value is not only required by metering but also as a logic of the method. So I would like to not execute this logic twice. 

I truly understand what you mean and why it is a good practice but for example, in `jenkins-idler` we had similar problem and we called `Recorder` directly.  For this reason, I decided to implement it in the same way for `now.`

internal/router/router.go
```diff
@@ -3,14 +3,14 @@ package router
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 // CreateAPIRouter is creating a router for the REST API of the Proxy.
 func CreateAPIRouter(api api.ProxyAPI) *httprouter.Router {
 	// Create router for API
 	proxyRouter := httprouter.New()
 	proxyRouter.GET("/api/info/:namespace", api.Info)
-	proxyRouter.Handler("GET", "/metrics", prometheus.Handler())
+	proxyRouter.Handler("GET", "/metrics", promhttp.Handler())
```
nurali-techie: Don't there is need to change package to work go build?
Here somewhere - https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/master/Gopkg.lock#L190

lordofthejars:
@chmouel any hint about this? https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/242#discussion_r184097375 I have updated the file for just in case.

Gopkg.lock
```diff
@@ -187,7 +187,10 @@
 
 [[projects]]
   name = "github.com/prometheus/client_golang"
-  packages = ["prometheus"]
+  packages = [
+    "prometheus",
+    "prometheus/promhttp"
+  ]
```
nurali-techie: @lordofthejars With humanly observation this change look fine to me.  'dep' tool auto-generate this .lock file.  Please try using it to avoid any future human mistake.  You can refer https://gist.github.com/subfuzion/12342599e26f5094e4e2d08e9d4ad50d or we can have bluejeans session to do it together.

internal/router/router.go
lordofthejars: I think that since we added the base package (prometheus client) the promhttp is automatically added as well (because it is a subpackage) so wedo not need to update anything as @chmouel noted.Also it compiles in CI env so it seems everything is correct. @nurali-techie are you ok on merging it?

Issue #243, opened 2018-04-27
Jenkins proxy page should show message explaining capacity is full


sthaha:
Honour capacity flag and display the information if the capacity is full. 

### Usecase

User is directed to https://jenkins.openshift.io and:
- Capacity Full flag is set
   1. display that the capacity is full 
   1. keep spinning/polling until jenkins can be unidled
   1. Once jenkins can be unidled, show the unidle screen and then the Jenkins.

- Capacity Full isn't set
   - Follow the existing unidle flow - show Unidling screen if idled else Jenkins

Refer: https://github.com/openshiftio/openshift.io/issues/3320

kishansagathiya:
@sthaha @containscafeine How about making `cluster capacity reached` an implementation detail?
In that proxy won't need to change a thing, All it needs to check is whether jenkins is up or not. Proxy wouldn't care if Jenkins is idle because the cluster has reached it's limit. WDYT?

kishansagathiya:
I guess we will have to tell why can't we unidle jenkins https://github.com/openshiftio/openshift.io/issues/3320#issuecomment-385367093

kishansagathiya:
Fixed with https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/248

PR #248, opened 2018-04-30
Issue #243 Show if Cluster Capacity is Reached

kishansagathiya:
Idler would be giving 503 if Cluster Tenant Resource Capacity is
Reached. Modified Unidle function to handle that

Now waiting template will show the message "Cluster Tenant resource
capacity for current cluster is full, please wait...", if
Cluster Capacity is reached

#243

### DIFF
--- a/internal/clients/idler.go
+++ b/internal/clients/idler.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/util"
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/util/logging"
-	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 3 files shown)


alien-ike:
### Ike Plugins (test-keeper)

internal/clients/idler.go
```diff
@@ -36,7 +36,7 @@ type clusterView struct {
 // IdlerService provides methods to talk to the idler client
 type IdlerService interface {
 	IsIdle(tenant string, openShiftAPIURL string) (bool, error)
-	UnIdle(tenant string, openShiftAPIURL string) error
+	UnIdle(tenant string, openShiftAPIURL string) (*int, error)
```
sthaha: We should be using an `int` instead of `int * ` here. 
I am not sure how a pointer to a const is even going to work.

kishansagathiya:
@sthaha @containscafeine 
What would be really nice for this is some UI magic such that the message that we see while jenkins is unidling changes dynamically i.e., If the ajax request that the page sent to proxy gives 503 change the message to `Cluster Tenant resource capacity for current cluster is full, please wait...`
and if it gives 202 it should change to `Jenkins has been idled. It is starting now, please wait...`

WDYT?

internal/proxy/proxy.go
```diff
@@ -318,12 +318,12 @@ func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request, r
...
 				}
 				if isIdle { //If jenkins is idled, return loading page and status 202
-					err = p.idler.UnIdle(ns, clusterURL)
+					code, err := p.idler.UnIdle(ns, clusterURL)
 					if err != nil {
 						p.HandleError(w, err, requestLogEntry)
 						return
 					}
-					err = p.processTemplate(w, ns, requestLogEntry)
+					err = p.processTemplate(w, ns, requestLogEntry, code)
```
containscafeine: we are not handling the error here

internal/proxy/proxy.go
```diff
@@ -339,7 +339,7 @@ func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request, r
 						cookie.Expires = time.Unix(0, 0)
 						http.SetCookie(w, cookie)
 					} else {
-						err = p.processTemplate(w, ns, requestLogEntry)
+						err = p.processTemplate(w, ns, requestLogEntry, http.StatusAccepted)
```
containscafeine: not handling here as well

kishansagathiya: Addressed in latest commit

containscafeine:
Nice work @kishansagathiya

PR #250, opened 2018-05-01
Issue #249 Make Redirection to Auth Temporary

### LINKED ISSUE #249: Change Redirect to Auth from Permanent(301) to Temporary(302 or 307)
local issue for https://github.com/openshiftio/openshift.io/issues/3392

301 is very problematic, It could create a link instead of directly redirecting.

https://github.com/kishansagathiya/fabric8-jenkins-proxy/blob/master/internal/proxy/proxy.go#L342

kishansagathiya:
Right now redirection to Auth for login is permanent (301).
Since permanent redirects are prolematic and might create a link
instead of directly redirecting. This should be changed to temporary
redirection(307), which is what this commit does.

Fixes #249

### DIFF
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -360,7 +360,7 @@ func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request, r
 	if needsAuth {
 		redirAuth := GetAuthURI(p.authURL, redirectURL.String())
 		requestLogEntry.Infof("Redirecting to auth: %s", redirAuth)
-		http.Redirect(w, r, redirAuth, 301)
+		http.Redirect(w, r, redirAuth, http.StatusTemporaryRedirect)
 	}
 	return
 }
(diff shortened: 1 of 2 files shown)


alien-ike:
### Ike Plugins (test-keeper)

internal/proxy/proxy.go
```diff
@@ -360,7 +360,7 @@ func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request, r
 	if needsAuth {
 		redirAuth := GetAuthURI(p.authURL, redirectURL.String())
 		requestLogEntry.Infof("Redirecting to auth: %s", redirAuth)
-		http.Redirect(w, r, redirAuth, 301)
+		http.Redirect(w, r, redirAuth, 307)
```
sthaha: isn't it better to use `http.StatusTemporaryRedirect` for clarity?

alien-ike:
Hey @kishansagathiya! It seems you tried to trigger `/ok-without-tests` command, but this will not have any effect due to insufficient permission. You have to be admin or requested reviewer or pull request approver, but not pull request creator for this command to take an effect.

PR #262, opened 2018-05-10
Issue #257: Using Go CORS Handler

### LINKED ISSUE #257: Make sure that CORS doesn't occur on Prod or Prod-preview
From what I understand, UI was never talking to Jenkins-proxy using Ajax calls. With https://github.com/fabric8-ui/fabric8-ui/pull/2889 merged it would do so.

Make sure that when openshift.io talks to jenkins.openshift.io we don't see CORS error. 
How are we making this checks? Does it have to do just with how applications are deployed?

Will jenkins.openshift.io and openshift.io be considered to have the same origin(i.e., openshift.io) or different? Do we have to change anything on the proxy code for this?

Tagging @sbose78 @sthaha

gastaldi:
This fixes #257 by reusing a CORS library

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -4,7 +4,10 @@
 [[projects]]
   branch = "master"
   name = "github.com/Azure/go-ansiterm"
-  packages = [".","winterm"]
+  packages = [
+    ".",
+    "winterm"
+  ]
   revision = "d6e3b3328b783f23731bc4d058875b0371ff8109"
 
 [[projects]]
@@ -51,7 +54,33 @@
 
 [[projects]]
   name = "github.com/docker/docker"
-  packages = ["api/types","api/types/blkiodev","api/types/container","api/types/filters","api/types/mount","api/types/network","api/types/registry","api/types/strslice","api/types/swarm","api/types/swarm/runtime","api/types/versions","opts","pkg/archive","pkg/fileutils","pkg/homedir","pkg/idtools","pkg/ioutils","pkg/jsonmessage","pkg/longpath","pkg/mount","pkg/pools","pkg/stdcopy","pkg/system","pkg/term","pkg/term/windows"]
+  packages = [
+    "api/types",
+    "api/types/blkiodev",
+    "api/types/container",
+    "api/types/filters",
+    "api/types/mount",
+    "api/types/network",
+    "api/types/registry",
+    "api/types/strslice",
+    "api/types/swarm",
+    "api/types/swarm/runtime",
+    "api/types/versions",
+    "opts",
+    "pkg/archive",
+    "pkg/fileutils",
+    "pkg/homedir",
+    "pkg/idtools",
+    "pkg/ioutils",
+    "pkg/jsonmessage",
+    "pkg/longpath",
+    "pkg/mount",
+    "pkg/pools",
+    "pkg/stdcopy",
+    "pkg/system",
+    "pkg/term",
+    "pkg/term/windows"
+  ]
   revision = "fe8aac6f5ae413a967adb0adad0b54abdfb825c4"
 
 [[projects]]
@@ -93,7 +122,17 @@
 [[projects]]
   branch = "master"
   name = "github.com/hashicorp/hcl"
-  packages = [".","hcl/ast","hcl/parser","hcl/scanner","hcl/strconv","hcl/token","json/parser","json/scanner","json/token"]
+  packages = [
+    ".",
+    "hcl/ast",
+    "hcl/parser",
+    "hcl/scanner",
+    "hcl/strconv",
+    "hcl/token",
+    "json/parser",
+    "json/scanner",
+    "json/token"
+  ]
   revision = "23c074d0eceb2b8a5bfdbb271ab780cde70f05a8"
 
 [[projects]]
@@ -122,7 +161,10 @@
 
 [[projects]]
   name = "github.com/lib/pq"
-  packages = [".","oid"]
+  packages = [
+    ".",
+    "oid"
+  ]
   revision = "83612a56d3dd153a94a629cd64925371c9adad78"
 
 [[projects]]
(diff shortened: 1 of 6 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @gastaldi for this contribution!



cmd/fabric8-jenkins-proxy/main.go
```diff
@@ -209,11 +211,16 @@ func startWorkers(ctx context.Context, wg *sync.WaitGroup, cancel context.Cancel
...
-func createProxyRouter(proxy *proxy.Proxy) *http.ServeMux {
+func createProxyRouter(proxy *proxy.Proxy) http.Handler {
 	proxyMux := http.NewServeMux()
 	proxyMux.HandleFunc("/", proxy.Handle)
 
-	return proxyMux
+	// cors.Default() setup the middleware with default options being
+	// all origins accepted with simple methods (GET, POST). See
+	// documentation in https://github.com/rs/cors for more options.
+	corsHandler := cors.Default().Handler(proxyMux)
```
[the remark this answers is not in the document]
gastaldi: Thank you, let me try that

Gopkg.lock
```diff
@@ -200,14 +259,29 @@
...
+  packages = [
+    ".",
+    "internal/util",
+    "nfs",
+    "xfs"
+  ]
+  revision = "8b1c2da0d56deffdbb9e48d4414b4e674bd8083e"
+
+[[projects]]
+  name = "github.com/rs/cors"
```
sthaha: Did you miss the committing Gopkg.toml?

Gopkg.lock
```diff
@@ -4,7 +4,10 @@
 [[projects]]
   branch = "master"
   name = "github.com/Azure/go-ansiterm"
-  packages = [".","winterm"]
+  packages = [
```
sthaha: seems to upgrade lot of unrelated packages. did you try `git add -p` to add only the relevant changes?

Gopkg.lock
```diff
@@ -187,7 +243,10 @@
 
 [[projects]]
   name = "github.com/prometheus/client_golang"
-  packages = ["prometheus"]
+  packages = [
+    "prometheus",
+    "prometheus/promhttp"
```
sthaha: not sure if this is needed for rs/cors

cmd/fabric8-jenkins-proxy/main.go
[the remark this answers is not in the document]
gastaldi: That doesn't work, I get
`cannot use cors.Default().Handler(proxyMux) (type http.Handler) as type *http.ServeMux in assignment: need type assertion`

Gopkg.lock
gastaldi: :man_shrugging:  This is the output from `make build`, it just indented the array I guess

Gopkg.lock
gastaldi: I didn't add that, `make build` did. Do I need to explicitly ignore that?

gastaldi:
As I am a Go newbie, I need some help writing tests for this (basically asserting that the CORS headers are sent correctly when requested), can anyone help me with this?

cmd/fabric8-jenkins-proxy/main.go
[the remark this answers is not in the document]
sthaha: I am confused about the the change to the ServeMux part which also changes the signature and thereby the meaning of createProxyRouter. The function does not return a `router`/`mux` anymore. I will have a closer look at this tomorrow.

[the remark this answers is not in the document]
gastaldi: The `http.Server(Handler:)` parameter expects a `Handler` implementation. See the example in https://github.com/rs/cors

gastaldi:
[test]

sbose78:
Tests added.
One of @sthaha @kishan @containscafeine  should review before we merge this :)

cmd/fabric8-jenkins-proxy/main.go
```diff
@@ -227,3 +214,25 @@ func setupSignalChannel(cancel context.CancelFunc) {
 		cancel()
 	}()
 }
+
+func getAPIServer(api api.ProxyAPI) *http.Server {
```
sthaha: nitpick: Golang getters just use the `Noun` instead of `GetNoun` see: https://golang.org/doc/effective_go.html#Getters

I don't think this is a getter since calling it creates a new  ApiServer all the time, hence in go-speak I would call it NewApiServer()

cmd/fabric8-jenkins-proxy/main.go
```diff
@@ -227,3 +214,25 @@ func setupSignalChannel(cancel context.CancelFunc) {
...
+	})
+	srv := &http.Server{
+		Addr:    apiRouterPort,
+		Handler: c.Handler(router.CreateAPIRouter(api)),
+	}
+	return srv
+}
+
+func getProxyServer(p *proxy.Proxy) *http.Server {
+	c := cors.New(cors.Options{
```
sthaha: I am not sure if we need cors for the reverse proxy. Could you please explain why this is needed?

sbose78: Just kept what was there in the previous commit I guess? :)

sthaha: @sbose78 aha, that is what I suspected. So in general, we don't need to CORS for reverse proxy side and only need it for the api server which is the part responsible for providing the `state` and the `unidle` api. We would also need to expose this through a route different route e.g. api.jenkins.openshift.io or jenkins.openshift.io/proxy/api/ since `/api/` IIUC is already an endpoint exposed by the actual jenkins.

sbose78: > We would also need to expose this through a route different route e.g. api.jenkins.openshift.io or jenkins.openshift.io/proxy/api/ since /api/ IIUC is already an endpoint exposed by the actual jenkins.

Absolutely agreed @sthaha . 
but please do check if it's OK to expose the API route, *in case* there was a good reason for not exposing it in the first place.

PR #268, opened 2018-05-17
Issue #267 Write Jenkins status API on a new server

### LINKED ISSUE #267: Create jenkins status api on a new server
Alternative of https://github.com/fabric8-services/fabric8-wit/issues/2095

kishansagathiya:
This commit creates a new server on :9092 and add an API at
`/api/status`, which would return current status of jenkinns and
trigger it unidling when the user's token_json(OSIO token) passed in
the query
```
Request: https://localhost:9092/api/status?token_json=$TOKEN_JSON
Response: { status: idled|starting|running }
```
Tested this to be working locally. 
@chmouel @sthaha Let me know if current approach look good to you. If so, I will move toward writing tests for this.

Part of https://github.com/openshiftio/openshift.io/issues/3356
Fixes #267

### DIFF
--- a/cmd/fabric8-jenkins-proxy/main.go
+++ b/cmd/fabric8-jenkins-proxy/main.go
@@ -7,6 +7,7 @@ import (
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/api"
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/clients"
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/configuration"
+	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/jenkinsapi"
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/proxy"
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/router"
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/storage"
@@ -30,6 +31,7 @@ const (
 	defaultStatsLoggingInterval = 5 * time.Minute
 	shutdownTimeout             = 5
 	apiRouterPort               = ":9091"
+	jenkinsAPIRouterPort        = ":9092"
 	proxyPort                   = ":8080"
 	profilerPort                = ":6060"
 )
@@ -166,6 +168,31 @@ func startWorkers(
 		}
 	}()
 
+	tenant := clients.NewTenant(config.GetTenantURL(), config.GetAuthToken())
+	idler := clients.NewIdler(config.GetIdlerURL())
+	jenkinsAPI := jenkinsapi.NewJenkinsAPI(&tenant, idler)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		srv := newJenkinsAPIServer(jenkinsAPI)
+
+		go func() {
+			mainLogger.Infof("Starting Jenkins Status API router on port %s", jenkinsAPIRouterPort)
+			listenAndServe(srv, cancel, config.GetHTTPSEnabled())
+		}()
+
+		for {
+			select {
+			case <-ctx.Done():
+				mainLogger.Infof("Shutting down Jenkins Status API router on port %s", jenkinsAPIRouterPort)
+				ctx, cancel := context.WithTimeout(ctx, shutdownTimeout*time.Second)
+				srv.Shutdown(ctx)
+				cancel()
+				return
+			}
+		}
+	}()
+
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
@@ -234,6 +261,17 @@ func newAPIServer(api api.ProxyAPI) *http.Server {
 	srv := &http.Server{
 		Addr:    apiRouterPort,
 		Handler: c.Handler(router.CreateAPIRouter(api)),
+	}
+	return srv
+}
+
+func newJenkinsAPIServer(jenkinsAPI jenkinsapi.JenkinsAPI) *http.Server {
+	c := cors.New(cors.Options{
+		AllowCredentials: true,
+	})
+	srv := &http.Server{
+		Addr:    jenkinsAPIRouterPort,
+		Handler: c.Handler(router.CreateJenkinsAPIRouter(jenkinsAPI)),
 	}
 	return srv
 }
--- a/internal/clients/idler.go
+++ b/internal/clients/idler.go
@@ -21,11 +21,12 @@ const (
 	OpenShiftAPIParam = "openshift_api_url"
 )
 
-type podState string
+// PodState represents status of the pod which could be running|starting|idled|
+type PodState string
 
 const (
 	// UnknownState used when the state isn't known, along with err
-	UnknownState podState = ""
+	UnknownState PodState = ""
 	// Running represents a running pod
 	Running = "running"
 	// Starting represents pod that is starting
@@ -41,28 +42,30 @@ type clusterView struct {
 	AppDNS string
 }
 
-// Jenkins pod state response structs
 // ErrorCode is an integer that clients to can use to compare errors
-type errorCode uint32
+type ErrorCode uint32
 
-type responseError struct {
-	Code        errorCode `json:"code"`
+// ResponseError contains and err description along with http status code
+type ResponseError struct {
+	Code        ErrorCode `json:"code"`
 	Description string    `json:"description"`
 }
 
-type jenkinsInfo struct {
-	State podState `json:"state"`
+// JenkinsInfo contains Jenkins pod state
+type JenkinsInfo struct {
+	State PodState `json:"state"`
 }
 
-type statusResponse struct {
-	Data   *jenkinsInfo    `json:"data,omitempty"`
-	Errors []responseError `json:"errors,omitempty"`
+// StatusResponse provides jenkins status and errors that occurred while making the status API call to idler
+type StatusResponse struct {
+	Data   *JenkinsInfo    `json:"data,omitempty"`
+	Errors []ResponseError `json:"errors,omitempty"`
 }
 
 // IdlerService provides methods to talk to the idler client
 type IdlerService interface {
 	UnIdle(tenant string, openShiftAPIURL string) (int, error)
-	State(tenant string, openShiftAPIURL string) (podState, error)
+	State(tenant string, openShiftAPIURL string) (PodState, error)
 	Clusters() (map[string]string, error)
 }
 
@@ -79,7 +82,7 @@ func NewIdler(url string) IdlerService {
 }
 
 // State returns the state of Jenkins instance for the specified tenant
-func (i *idler) State(tenant string, openShiftAPIURL string) (podState, error) {
+func (i *idler) State(tenant string, openShiftAPIURL string) (PodState, error) {
 	namespace := tenant
 	if !strings.HasSuffix(tenant, namespaceSuffix) {
 		namespace = tenant + namespaceSuffix
@@ -114,7 +117,7 @@ func (i *idler) State(tenant string, openShiftAPIURL string) (podState, error) {
 		return UnknownState, err
 	}
 
-	sr := &statusResponse{}
+	sr := &StatusResponse{}
 	err = json.Unmarshal(body, sr)
 	if err != nil {
 		return UnknownState, err
@@ -202,7 +205,7 @@ func (i *idler) Clusters() (map[string]string, error) {
 	return clusters, nil
 }
 
-func (e responseError) Error() string {
+func (e ResponseError) Error() string {
 	return fmt.Sprintf("%d: %s", e.Code, e.Description)
 }
 
--- a/internal/clients/tenant.go
+++ b/internal/clients/tenant.go
@@ -9,6 +9,8 @@ import (
 
 	"errors"
 
+	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/configuration"
+	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/util"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -27,6 +29,12 @@ func NewTenant(tenantServiceURL string, authToken string) Tenant {
 		client:           &http.Client{},
 		logger:           logger,
 	}
+}
+
+// TenantService is contains methods that makes calls to tenant APIs
+type TenantService interface {
+	GetTenantInfo(tenantID string) (TenantInfo, error)
+	GetNamespace(accessToken string) (Namespace, error)
 }
 
 // Tenant is a simple client for fabric8-tenant.
@@ -116,7 +124,7 @@ func (t Tenant) GetTenantInfo(tenantID string) (ti TenantInfo, err error) {
 }
 
 // GetNamespaceByType searches tenant namespaces for a given type.
-func (t Tenant) GetNamespaceByType(ti TenantInfo, typ string) (r Namespace, err error) {
+func GetNamespaceByType(ti TenantInfo, typ string) (r Namespace, err error) {
 	for i := 0; i < len(ti.Data.Attributes.Namespaces); i++ {
 		n := ti.Data.Attributes.Namespaces[i]
 		if n.Type == typ {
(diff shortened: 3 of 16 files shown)


alien-ike:
### Ike Plugins (test-keeper)

internal/clients/idler.go
```diff
@@ -32,9 +32,26 @@ type clusterView struct {
 	AppDNS string
 }
 
+type responseError struct {
+	Code        errorCode `json:"code"`
+	Description string    `json:"description"`
+}
+
+type jenkinsInfo struct {
+	State string `json:"state"`
```
sthaha: the State needs to be a type like the one here so that it is type-safe.

https://github.com/sthaha/fabric8-jenkins-proxy/commit/97d7357d13e5543d9878d6018c4e09c9a90ee060

internal/clients/idler.go
```diff
@@ -92,6 +109,47 @@ func (i *idler) IsIdle(tenant string, openShiftAPIURL string) (bool, error) {
...
+		return "", err
+	}
+
+	s := &statusResponse{}
+	err = json.Unmarshal(body, s)
+	if err != nil {
+		return "", err
+	}
+
+	log.Debugf("Jenkins status is (%t) in %s", s.Data.State, namespace)
```
sthaha: Like in https://github.com/sthaha/fabric8-jenkins-proxy/commit/97d7357d13e5543d9878d6018c4e09c9a90ee060

It need to check if the Error is empty and return the first error

internal/statusapi/statusapi.go
```diff
@@ -0,0 +1,107 @@
...
+		tenant: tenant,
+		idler:  idler,
+	}
+}
+
+// Status returns the Jenkins status for the current user
+func (api *jenkinsStatusAPIImpl) Status(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+
+	log.Infof("Status API ran")
+	tokenJSON, ok := r.URL.Query()["token_json"]
```
sthaha: this should be a bearer token in a header instead of a query parameter?

internal/statusapi/statusapi.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/clients"
+	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/configuration"
+	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/proxy"
+	"github.com/julienschmidt/httprouter"
+	log "github.com/sirupsen/logrus"
+)
+
+//JenkinsStatusAPI contains API to check whether Jenkins for the current user is idle|running|starting
+type JenkinsStatusAPI interface {
+	Status(w http.ResponseWriter, r *http.Request, ps httprouter.Params)
```
sthaha: we don't need a ps param here since it is unused.

internal/statusapi/statusapi.go
```diff
@@ -0,0 +1,107 @@
...
+	idler  clients.IdlerService
+}
+
+// JenkinsStatus represents whether jenkins is running|starting|idle
+type JenkinsStatus struct {
+	Status string `json:"status"`
+}
+
+// NewJenkinsStatusAPI creates a new instance of JenkinsStatusAPI
+func NewJenkinsStatusAPI(tenant *clients.Tenant, idler clients.IdlerService) JenkinsStatusAPI {
```
sthaha: shouldn't the `idler` be a pointer as well?

internal/statusapi/statusapi.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	log.Infof("Found token info in the query. Namespace is %s and clusterURL is %s", namespace.Name, namespace.ClusterURL)
+
+	status, err := api.idler.Status(namespace.Name, namespace.ClusterURL)
+	if err != nil {
+		log.Error(err)
+		return
+	}
+
+	if status == "idled" {
```
sthaha: instead of harded `"idled"` plesae consider using a constant like in https://github.com/sthaha/fabric8-jenkins-proxy/commit/97d7357d13e5543d9878d6018c4e09c9a90ee060

sthaha: and it should be `status != clients.Running`

internal/statusapi/statusapi.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+	}
+
+	json.NewEncoder(w).Encode(JenkinsStatus{
+		Status: status,
+	})
+
+}
+
+func processToken(tokenData string, tenant *clients.Tenant) (*clients.Namespace, error) {
```
sthaha: how about calling this `namespaceForToken` ?

cmd/fabric8-jenkins-proxy/main.go
```diff
@@ -238,6 +265,17 @@ func newAPIServer(api api.ProxyAPI) *http.Server {
 	return srv
 }
 
+func newStatusAPIServer(statusapi statusapi.JenkinsStatusAPI) *http.Server {
+	c := cors.New(cors.Options{
+		AllowCredentials: true,
+	})
+	srv := &http.Server{
+		Addr:    jenkinsStatusAPIRouterPort,
+		Handler: c.Handler(router.CreateJenkinsStatusAPIRouter(statusapi)),
```
sthaha: It may be better to create a middleware that does the authentication so that all api need not worry about auth as it will be taken care by the middleware.

cmd/fabric8-jenkins-proxy/main.go
```diff
@@ -166,6 +168,31 @@ func startWorkers(
 		}
 	}()
 
+	tenant := clients.NewTenant(config.GetTenantURL(), config.GetAuthToken())
+	idler := clients.NewIdler(config.GetIdlerURL())
+	statusapi := statusapi.NewJenkinsStatusAPI(&tenant, idler)
```
sthaha: why not pass idler as a pointer as well to keep the interface consistent?

internal/statusapi/statusapi.go
```diff
@@ -0,0 +1,107 @@
...
+	log.Infof("Status API ran")
+	tokenJSON, ok := r.URL.Query()["token_json"]
+	if !ok {
+		log.Errorf("Couldn't find token_json in the query")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	namespace, err := processToken(tokenJSON[0], api.tenant)
+	if err != nil {
+		log.Error(err)
```
sthaha: what response will this write?

internal/statusapi/statusapi.go
```diff
@@ -0,0 +1,107 @@
...
+		return
+	}
+	log.Infof("Found token info in the query. Namespace is %s and clusterURL is %s", namespace.Name, namespace.ClusterURL)
+
+	status, err := api.idler.Status(namespace.Name, namespace.ClusterURL)
+	if err != nil {
+		log.Error(err)
```
sthaha: same, this needs to write a response - internal error and return an error: struct just like the idler status api.

internal/statusapi/statusapi.go
```diff
@@ -0,0 +1,107 @@
...
+
+// JenkinsStatusAPIImpl implements JenkinsStatusAPI
+type jenkinsStatusAPIImpl struct {
+	tenant *clients.Tenant
+	idler  clients.IdlerService
+}
+
+// JenkinsStatus represents whether jenkins is running|starting|idle
+type JenkinsStatus struct {
+	Status string `json:"status"`
```
sthaha: could you please consider modelling this after the idler statusResponse struct? so that it provdes `data` and `errors` ?

internal/statusapi/statusapi.go
```diff
@@ -0,0 +1,107 @@
...
+// NewJenkinsStatusAPI creates a new instance of JenkinsStatusAPI
+func NewJenkinsStatusAPI(tenant *clients.Tenant, idler clients.IdlerService) JenkinsStatusAPI {
+	return &jenkinsStatusAPIImpl{
+		tenant: tenant,
+		idler:  idler,
+	}
+}
+
+// Status returns the Jenkins status for the current user
+func (api *jenkinsStatusAPIImpl) Status(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
```
sthaha: we are not using ps in the function, are we?

internal/statusapi/statusapi.go
kishansagathiya: url?token_json is easy to use compared to putting it in the authorization header. Auth service uses it the same way. When you login to osio you will see openshift.io?token_json=$TOKEN_JSON.

internal/statusapi/statusapi.go
kishansagathiya: ps.ByName was giving errors. Getting rid of ps

kishansagathiya: oops can't get rid of ps even if I don't need it. Otherwise router won't accept it as handle. Must be due to some httprouter requirement

sthaha: right, isn't golang convention to use `_` in that case?

internal/statusapi/statusapi.go
sthaha: in that case, shouldn't it be tenant interface?

internal/statusapi/statusapi.go
sthaha: that is the redirection part which then requires that token to be preserved somehow so that UI can use the token while making ajax calls. If you look at the UI code, you will find that they are using Auth header instead of query param ?token_json. Bottom line is that it is easier to set a header than manipulate the query.

sthaha: @kishansagathiya see: https://github.com/fabric8-ui/fabric8-ui/pull/2895/files look for `headers.set('Authorization', 'Bearer ' + token);`

internal/statusapi/statusapi.go
kishansagathiya: there isnt one

internal/statusapi/statusapi.go
kishansagathiya: Noted. Will do

internal/statusapi/statusapi.go
kishansagathiya: Done

internal/statusapi/statusapi.go
kishansagathiya: Done

internal/statusapi/statusapi.go
kishansagathiya: done

internal/statusapi/statusapi.go
kishansagathiya: done

internal/statusapi/statusapi.go
kishansagathiya: done

internal/statusapi/statusapi.go
kishansagathiya: Done with https://github.com/fabric8-services/fabric8-jenkins-proxy/pull/268/commits/3f60200b263f94df8aecbc9448fc64e38bc11b6e

internal/router/router.go
```diff
@@ -18,6 +19,14 @@ func CreateAPIRouter(api api.ProxyAPI) *httprouter.Router {
 	return proxyRouter
 }
 
+// CreateJenkinsAPIRouter is creating a router for the REST API of the Proxy.
+func CreateJenkinsAPIRouter(jenkinsAPI jenkinsapi.JenkinsAPI) *httprouter.Router {
+	// Create router for API
+	jenkinsAPIRouter := httprouter.New()
+	jenkinsAPIRouter.GET("/api/jenkins/start", jenkinsAPI.Start)
```
sthaha: this should be `POST` instead of a `get` since we are changing the state of a resource.

internal/jenkinsapi/jenkinsapi.go
```diff
@@ -0,0 +1,114 @@
...
+	ti, err := tenant.GetTenantInfo(uid)
+	if err != nil {
+		return nil, err
+	}
+	namespace, err := tenant.GetNamespaceByType(ti, "jenkins")
+	if err != nil {
+		return nil, err
+	}
+
+	return &namespace, nil
```
sthaha: this makes me think that the method must be implemented by `client.Tenant` instead of a function that receives `ti`


e.g. 
```go
namspace, err := tenant.NamespaceForToken(accessToken)
```
nitpick: the golang idiom is not to use `get` which is why I suggested NamespaceForToken.

cmd/fabric8-jenkins-proxy/main.go
```diff
@@ -238,6 +265,17 @@ func newAPIServer(api api.ProxyAPI) *http.Server {
 	return srv
 }
 
+func newJenkinsAPIServer(jenkinsAPI jenkinsapi.JenkinsAPI) *http.Server {
+	c := cors.New(cors.Options{
```
sthaha: should we remove the cors middleware from the other endpoints?

internal/jenkinsapi/jenkinsapi.go
kishansagathiya: That's actually more appropriate. Will move `NamespaceForToken` to tenant client. I don't think tenant client is trying to be consistent with tenant service APIs. 

cc @chmouel

internal/clients/tenant.go
```diff
@@ -29,6 +31,12 @@ func NewTenant(tenantServiceURL string, authToken string) Tenant {
 	}
 }
 
+// TenantService is contains methods that makes calls to tenant APIs
+type TenantService interface {
+	GetTenantInfo(tenantID string) (ti TenantInfo, err error)
```
sthaha: nitpick: can remove `ti` and `err`

internal/clients/tenant.go
```diff
@@ -128,3 +136,32 @@ func (t Tenant) GetNamespaceByType(ti TenantInfo, typ string) (r Namespace, err
 	err = fmt.Errorf("could not find tenant %s Jenkins namespace", ti.Data.Attributes.Email)
 	return
 }
+
+// GetNamespace gets namespace given appropriate accessToken
+func (t Tenant) GetNamespace(accessToken string) (*Namespace, error) {
```
sthaha: why *Namespace when GetNamespaceByType returns value? 
IMHO, it should be `func (t Tenant) GetNamespace(accessToken string) (Namespace, error)` and `return namespace, nil`

kishansagathiya: `*Namespace, error` allows to return `nil, err`

internal/clients/tenant.go
kishansagathiya: This is as per old implementation. If we are to do this, let's do it with next PR.

internal/clients/tenant.go
sthaha: yes, so what does that save the caller? 
now you have to check the `error` and then check if the object is nil? 
heard of https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare ? https://en.wikipedia.org/wiki/Tony_Hoare

kishansagathiya: Work for the caller is same in both cases.

sthaha: It isn't .. for a pointer, they have to check for nil every single time it is accessed. 
```go
package main

import "fmt"

type person struct {
	name string
}

func value() person {
	return person{}
}

func pointer() *person {
	return nil
}

func main() {

	v := value()
	fmt.Println(v.name)

	p := pointer()
	fmt.Println(p.name)  // ### crash ###
}
```

sthaha: May be a confirmation bias still - https://www.reddit.com/r/golang/comments/2xmnvs/returning_nil_for_a_struct/cp1h15d

internal/clients/tenant.go
sthaha: do you mean we need a new PR for  removing those vars `ti` and `err` ? 
i.e. to make it `GetTenantInfo(tenantID string) (TenantInfo, error)` ?

PR #271, opened 2018-05-18
Fix #270 Use new idler status api to check pod state

### LINKED ISSUE #270: Deprecate idler.IsIdle method in favour of the new status api
Proxy uses `! idler.IsIdle()` to check if pods are `running` but that is wrong as `IsIdle` can return `false` if  pods are just `starting` or `terminating`.  Hence lets use the new api in idler `status` to check if the Pods are actually running.

sthaha:
Proxy uses `!idler.IsIdle()` to check if pods are running but that is
wrong as IsIdle can return `false` if pods are just `starting` or
`terminating`. Hence lets use the new api in idler status to check if the
Pods are actually running.

### DIFF
--- a/internal/clients/idler.go
+++ b/internal/clients/idler.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/util"
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/util/logging"
@@ -20,10 +21,18 @@ const (
 	OpenShiftAPIParam = "openshift_api_url"
 )
 
-// status represent status of the service true if idle, false if unidle.
-type status struct {
-	IsIdle bool `json:"is_idle"`
-}
+type podState string
+
+const (
+	// UnknownState used when the state isn't known, along with err
+	UnknownState podState = ""
+	// Running represents a running pod
+	Running = "running"
+	// Starting represents pod that is starting
+	Starting = "starting"
+	// Idled represents pod that is idled
+	Idled = "idled"
+)
 
 // clusterView is a view of the cluster topology which only includes the OpenShift API URL and the application DNS for this
 // cluster.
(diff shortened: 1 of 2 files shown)


alien-ike:
### Ike Plugins (test-keeper)

alien-ike:
Hey @sthaha! It seems you tried to trigger `/ok-without-tests` command, but this will not have any effect due to insufficient permission. You have to be admin or requested reviewer or pull request approver, but not pull request creator for this command to take an effect.

internal/clients/idler.go
```diff
@@ -32,10 +41,28 @@ type clusterView struct {
...
+type statusResponse struct {
+	Data   *jenkinsInfo    `json:"data,omitempty"`
+	Errors []responseError `json:"errors,omitempty"`
+}
+
 // IdlerService provides methods to talk to the idler client
 type IdlerService interface {
-	IsIdle(tenant string, openShiftAPIURL string) (bool, error)
 	UnIdle(tenant string, openShiftAPIURL string) (int, error)
+	State(tenant string, openShiftAPIURL string) (podState, error)
```
kishansagathiya: Let's keep this consistent with idler APIs. I think there it is `Status`, isn't it?

internal/clients/idler.go
```diff
@@ -51,45 +78,52 @@ func NewIdler(url string) IdlerService {
 	}
 }
 
-// IsIdle returns true if the Jenkins instance for the specified tenant is idled, false otherwise.
-func (i *idler) IsIdle(tenant string, openShiftAPIURL string) (bool, error) {
+// State returns the state of Jenkins instance for the specified tenant
+func (i *idler) State(tenant string, openShiftAPIURL string) (podState, error) {
```
kishansagathiya: What's the incentive behind using podState instead of string?

internal/clients/idler.go
```diff
@@ -20,10 +21,18 @@ const (
...
-type status struct {
-	IsIdle bool `json:"is_idle"`
-}
+type podState string
+
+const (
+	// UnknownState used when the state isn't known, along with err
+	UnknownState podState = ""
+	// Running represents a running pod
+	Running = "running"
```
kishansagathiya: Why is `running` not a podState? Can I use podState and string interchangeably?

internal/clients/idler.go
sthaha: yes, but status is meant to return the entire datastructure related to status which could include many attributes not just the state of a pod.

internal/clients/idler.go
```diff
@@ -51,45 +78,52 @@ func NewIdler(url string) IdlerService {
...
+		return UnknownState, err
 	}
 
-	log.Debugf("Jenkins is idle (%t) in %s", s.IsIdle, namespace)
+	if len(sr.Errors) != 0 {
+		return UnknownState, sr.Errors[0]
 
-	return s.IsIdle, nil
+	}
+
```
kishansagathiya: Let's put a log here?

internal/clients/idler.go
sthaha: It is a string but type aliased as podState so that golang can type check it e.g.
the code below produces error. 

```go
package main

type Color string

const (
	Red Color = "red"
)

type Name string

const (
	Foobar Name = "foobar"
)

func main() {
	var x Color = Foobar // ERROR
}

```

internal/clients/idler.go
sthaha: Running is a `podState` because the const started with declaring the first const `UnknownState` as a `podState`

internal/clients/idler.go
sthaha: ack .. will submit a correction

kishansagathiya:
`/ok-without-tests`

alien-ike:
Hey @sthaha! It seems you tried to trigger `/ok-without-tests` command, but this will not have any effect due to insufficient permission. You have to be admin or requested reviewer or pull request approver, but not pull request creator for this command to take an effect.

PR #273, opened 2018-05-22
Fix #256 send only one request to get the status of Jenkins pod

### LINKED ISSUE #256: Jenkins Proxy Loading page sends 2 XHR instead of one while waiting for jenkins to get unidled
## Steps to reproduce

1. Wait for jenkins to be idled
2. goto https:jenkins.openshift.io
3. Watch network requests to proxy

## Expected:
* must see only one request to check if jenkins is up and running

## Actual
* You would find two request being sent to proxy

## Probable cause & solution:
`index.html` makes two calls  - `check()` and then `setTimeout`. you would find that `check` already does a `setTimeout` so deleting the `setTimeout` after `check` may solve this

sthaha:
The patch also delays reloading the page so that Jenkins pods
get some time after it boots to initialise properly.

### DIFF
--- a/static/html/index.html
+++ b/static/html/index.html
@@ -34,13 +34,14 @@
               200: function() {
                 console.log("Got 200, reloading...")
                 showStatus("Starting Jenkins", "Jenkins is up and running; loading..." )
-                window.location.reload(true)
-                },
+                // wait a bit longer for jenkins to boot properly after idle
+                setTimeout(function(){ window.location.reload(true) }, timeout)
+              },
               202: function(){
                 console.log("Got 202, waiting along..")
                 showStatus("Starting Jenkins", "Jenkins is currently idled. Please wait while we start it...")
                 setTimeout(check, timeout)
-                },
+              },
               503: function(){
                 console.log("Got 503, Cluster resources capacity is full. waiting along..")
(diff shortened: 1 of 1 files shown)


kishansagathiya:
@sthaha 
Not sure why we need this. 200 is seen only when jenkins is running, right? Then, on 200 can't we just proxy to jenkins?

kishansagathiya:
https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/d06f9df0c86558decf509886527023004328e21c/internal/proxy/ui_requests.go#L141-L158

If we get StatusOK, it mean jenkins is running. What's the point of waiting if it is running?

sthaha:
@kishansagathiya 
> If we get StatusOK, it mean jenkins is running. What's the point of waiting if it is running?

Because running doesn't mean jenkins is really up and running :), It only means the login was successful and at which point, the numerous plugins are/maybe doing something CPU intensive (as observed) and Jenkins fails to load properly. 

Note, this delay (loading page ...) is applied only when jenkins is idle.

kishansagathiya:
@sthaha 
So, If Jenkins pod is up, doesn't mean Jenkins is running?

PR #275, opened 2018-05-22
[DNM] Issue #274 Check cookies first

### LINKED ISSUE #274: Check for cookies first and if no cookies then look for token_json

kishansagathiya:
While proxying using UI, first check if we find JSESSIONID cookie or
JenkinsIdled cookie and if we don't then look for token_json in the URL

Fixes #274

### DIFF
--- a/internal/proxy/ui_requests.go
+++ b/internal/proxy/ui_requests.go
@@ -4,13 +4,14 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/clients"
-	"github.com/satori/go.uuid"
-	log "github.com/sirupsen/logrus"
 	"net/http"
 	"net/url"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


alien-ike:
### Ike Plugins (test-keeper)

kishansagathiya:
Need to check if this is breaking anything

internal/proxy/ui_requests.go
```diff
@@ -4,13 +4,14 @@ import (
...
 	"errors"
 	"fmt"
-	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/clients"
-	"github.com/satori/go.uuid"
-	log "github.com/sirupsen/logrus"
 	"net/http"
 	"net/url"
 	"strings"
 	"time"
+
```
sthaha: does go fmt agree with this?

kishansagathiya: That is done by go fmt. I didnt touch it.

PR #277, opened 2018-05-22
WIP: Fix #TBD Ensure jenkins starts even if there is a session cookie

sthaha:


alien-ike:
### Ike Plugins (test-keeper)

internal/proxy/ui_requests.go
```diff
@@ -116,17 +121,59 @@ func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request, r
...
+						p.HandleError(w, err, requestLogEntry)
+						return
+					}
+					p.recordStatistics(pci.NS, time.Now().Unix(), 0) //FIXME - maybe do this at the beginning?
+
+				} else {
+					r.Host = pci.Route //Configure proxy upstream
+					r.URL.Host = pci.Route
+					r.URL.Scheme = pci.Scheme
+					needsAuth = false //user is probably logged in, do not redirect
```
kishansagathiya: Should get rid of `probably` from the comment

internal/proxy/ui_requests.go
```diff
@@ -224,6 +271,27 @@ func (p *Proxy) loginJenkins(pci CacheItem, osoToken string, requestLogEntry *lo
...
+	return strings.HasPrefix(c.Name, SessionCookie)
+}
+
+func isIdledCookie(c *http.Cookie) bool {
+	return c.Name == CookieJenkinsIdled
+}
+
+func anyCookie(_ *http.Cookie) bool {
+	return true
+}
```
kishansagathiya: It seems to me that current method will clear all cookies set by jenkins.openshift.io, is that right?

internal/proxy/ui_requests.go
```diff
@@ -116,17 +121,59 @@ func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request, r
...
+					}
+
+					if code == http.StatusOK {
+						code = http.StatusAccepted
+					} else if code != http.StatusServiceUnavailable {
+						p.HandleError(w, fmt.Errorf("Failed to send unidle request using fabric8-jenkins-idler"), requestLogEntry)
+					}
+
+					w.WriteHeader(code)
+					clearCookiesMatching(w, r, isSessionCookie)
```
kishansagathiya: I think JSESSIONID is the only cookie that is being created multiple times, right?
JenkinsIdled cannot be, because it will always have the same name.

internal/proxy/ui_requests.go
sthaha: all cookies will be cleared **iff** clearCookiesMatching is passed anyCookie. I am thinking of doing this when `token_json` is passed in the query and jenkins is idled .. I will revisit this thought since I think there could be jenkins-preference cookie that may still be valid.

internal/proxy/ui_requests.go
sthaha: ack ... :)

sthaha:
Closing it as the patch got merged in a different PR

PR #282, opened 2018-05-28
Fix #280 Unidle Jenkins only if needed

### LINKED ISSUE #280: Proxy bugs idler a lot!
Every time someone calls `jenkins.openshift.io`, proxy jumps into action and tried to un-idle the jenkins pod. Here is how it does it -

1. Once token_json is set by auth, it un-idles, sets the cookies, and redirects to self to remove token_json from the URL -
https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/0c4f75bfc810aaf18291645da3aad8d946fc9c8f/internal/proxy/ui_requests.go#L64-L76

2. After being redirected, instantaneously, proxy finds the cookies and un-idles again and calls `processTemplate()` -
https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/0c4f75bfc810aaf18291645da3aad8d946fc9c8f/internal/proxy/ui_requests.go#L129-L160

3. The template sends a request to self yet again, which means calling proxy again and calling idler again -
https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/0c4f75bfc810aaf18291645da3aad8d946fc9c8f/static/html/index.html#L29-L32

This means that's 3 (or maybe more undiscovered) un-idle requests from proxy. This sort of messes up the logs, and _probably_ causes max retries to be reached in idler making it unreliable for un-idling.

sthaha:
This patch refactors un-idling to proxy's startJenkins function
which unildes jenkins only if it the state is "idled"

### DIFF
--- a/internal/proxy/ui_requests.go
+++ b/internal/proxy/ui_requests.go
@@ -57,26 +57,29 @@ func (p *Proxy) handleJenkinsUIRequest(w http.ResponseWriter, r *http.Request, r
 		nsLogger := requestLogEntry.WithFields(log.Fields{"ns": ns, "cluster": clusterURL})
 		nsLogger.Debug("Found token info in query")
 
-		state, err := p.idler.State(ns, clusterURL)
+		// we don't care about code here since only the state of jenkins pod
+		// running or not is what is relevant
+
+		state, _, err := p.startJenkins(ns, clusterURL)
 		if err != nil {
 			p.HandleError(w, err, requestLogEntry)
 			return
 		}
 
-		//Break the process if the Jenkins is idled, set a cookie and redirect to self
 		if state != clients.Running {
-			_, err := p.idler.UnIdle(ns, clusterURL)
-			if err != nil {
-				p.HandleError(w, err, requestLogEntry)
-				return
-			}
+			// Break the process if Jenkins isn't running.
 
+			// Set "idled" cookie to indicate that jenkins is idled
+			// also cache the ns & cluster for faster lookup next time
 			p.setIdledCookie(w, pci)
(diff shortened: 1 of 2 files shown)


internal/proxy/ui_requests.go
```diff
@@ -270,3 +271,31 @@ func (p *Proxy) processToken(tokenData []byte, requestLogEntry *log.Entry) (pci
...
+func (p *Proxy) startJenkins(ns, clusterURL string) (state clients.PodState, code int, err error) {
+	// Assume pods are starting and unidle only if it is in "idled" state
+	code = http.StatusAccepted
+
+	state, err = p.idler.State(ns, clusterURL)
+	if err != nil {
+		return
+	}
+
+	if state == clients.Idled {
```
kishansagathiya: I remember you saying that even if jenkins is idling then also we have three states
Running -> Intermediate -> Idled. Is that right?

If the state is Intermediate, it would be worth knowing if Jenkins is idling or un-idling. But let's do that as a different PR, WDYT?

sthaha: @kishansagathiya at present idler does not give us that information. We need to build that logic into idler which should return state like "starting", "running", "idling", "idled" which at the moment, it doesn't and thus can't be built into proxy.

kishansagathiya:
[ test ]

internal/proxy/ui_requests.go
sthaha: @kishansagathiya  https://github.com/fabric8-services/fabric8-jenkins-idler/issues/246

sthaha:
[test]

PR #292, opened 2018-06-06
Issue 290 Fix buggy ProcessBuffer

kishansagathiya:


internal/proxy/github.go
```diff
@@ -220,9 +222,7 @@ func (p *Proxy) getUserWithRetry(repositoryCloneURL string, logEntry *log.Entry,
 
 	}
 
-	// Last chance
-	return p.getUser(repositoryCloneURL, logEntry)
-
+	return clients.Namespace{}, errors.New("Couldn't get User, even after retrying " + string(retry) + " times")
```
chmouel: I don't know the code but is clients.Namespace{} return the same as p.getUser?

kishansagathiya: Yup, but this is basically one more retry. If what we want to do is just that, why don't we just increase the retry parameter by 1? It gives takes parameter `retry` and retries for `retry + 1`. Inappropriate.

sthaha: I agree with solving the retry count issue but the error must the be same error as `getUser` How about - 

```go
for i := 1; i < retry; i++ {
  if ns, err := p.getUser(repositoryCloneURL, logEntry); err == nil {
        return ns, nil
  }
  time.Sleep(1*time.Second)
}
return  p.getUser(repositoryCloneURL, logEntry);

Comment (kishansagathiya) [reply internal/proxy/github.go]: @sthaha code piece which you are suggesting just makes the code compact and doesn't solve the retry count issue. Anyway, retry count issue will not create any problem, so I will change it to what you are suggesting.

File: internal/proxy/github.go [hunk 2]
```diff
@@ -53,13 +53,14 @@ func (p *Proxy) handleGitHubRequest(w http.ResponseWriter, r *http.Request, requ
...
-	ns = namespace.Name
-	clusterURL := namespace.ClusterURL
-	requestLogEntry.WithFields(log.Fields{"ns": ns, "cluster": clusterURL, "repository": gh.Repository.CloneURL}).Info("Processing GitHub request ")
 	if err != nil {
 		p.HandleError(w, err, requestLogEntry)
 		return
 	}
+	ns = namespace.Name
+	clusterURL := namespace.ClusterURL
+	requestLogEntry.WithFields(log.Fields{"ns": ns, "cluster": clusterURL, "repository": gh.Repository.CloneURL}).Info("Processing GitHub request ")
```
sthaha: since you have the details of the ns and clusterURL here, it better to create a new logger with those fields added and use that to log all subsequent messages. 
e.g. 
```go 
nsLogger := requestLogEntry.WithFields(...)

# else where
nsLogger.Infof(...)
```

internal/proxy/github.go
```diff
@@ -175,7 +177,11 @@ func (p *Proxy) ProcessBuffer() {
 								break
 							}
 
-							if resp.StatusCode != 200 && resp.StatusCode != 404 { //Retry later if the response is not 200
+							if resp.StatusCode == 200 {
+								log.WithField("ns", ns).Infof(fmt.Sprintf("Request for %s to %s forwarded.", ns, req.Host))
```
sthaha: why fmt.Sprintf? wouldn't Infof suffice? Also the ns is already logged as a Field so we don't have to repeat and use %q instead of %s so that we know if Host is "" by any chance.

Also can't we return here?

internal/proxy/github.go
```diff
@@ -175,7 +177,11 @@ func (p *Proxy) ProcessBuffer() {
...
+							} else if resp.StatusCode == 404 || resp.StatusCode == 400 { //400 - missing payload
+								log.Warn(fmt.Sprintf("Got status %s after retrying request on %s, throwing away the request", resp.Status, req.URL))
```
sthaha: same, please use Warnf instead of Warn(fmt.Sprintf( .. ))

internal/proxy/github.go
```diff
@@ -175,7 +177,11 @@ func (p *Proxy) ProcessBuffer() {
...
+							} else { //Retry later if the response is not 200
```
sthaha: The comment isn't true anymore

internal/proxy/github.go
```diff
@@ -175,7 +177,11 @@ func (p *Proxy) ProcessBuffer() {
...
 								log.Error(fmt.Sprintf("Got status %s after retrying request on %s", resp.Status, req.URL))
 								errs := p.storageService.IncrementRequestRetry(&r)
 								if len(errs) > 0 {
```
sthaha: if is just not needed

internal/proxy/github.go
```diff
@@ -184,14 +190,12 @@ func (p *Proxy) ProcessBuffer() {
...
 								break
-							} else if resp.StatusCode == 404 || resp.StatusCode == 400 { //400 - missing payload
-								log.Warn(fmt.Sprintf("Got status %s after retrying request on %s, throwing away the request", resp.Status, req.URL))
 							}
 
-							log.WithField("ns", ns).Infof(fmt.Sprintf("Request for %s to %s forwarded.", ns, req.Host))
 						}
 
-						//Delete request if we tried too many times or the replay was successful
+						// Deleting request since we tried too many times or the replay was successful
```
sthaha: wouldn't this part be executed even if r.Retries < Max ?

internal/proxy/proxy.go
```diff
@@ -73,7 +73,7 @@ func NewProxy(
 		tenant:           tenant,
 		wit:              wit,
 		idler:            idler,
-		bufferCheckSleep: 30,
+		bufferCheckSleep: 30 * time.Second,
```
sthaha: shouldn't the type be also changed to time.Duration ?

internal/proxy/proxy_test.go
```diff
@@ -40,5 +40,5 @@ func TestGetUserWithRetry(t *testing.T) {
 
 	_, err := p.getUserWithRetry("http://test", logEntry, numberofretry)
 	assert.Error(t, err, "Faker")
-	assert.Equal(t, numberofretry+1, wit.testCounter)
+	assert.Equal(t, numberofretry, wit.testCounter)
```
sthaha: +1 :)

internal/proxy/github.go
sthaha: if the above is a return, then this can be changed to `if` instead of `else if`. Also, can't we return?

internal/proxy/github.go
kishansagathiya: ns and clusterURL are being logged only once. Why should I create a separate logger just for that?

internal/proxy/github.go
kishansagathiya: ProcessBuffer keeps checking requests and keeps processing it. It never returns.

internal/proxy/github.go
kishansagathiya: Yes, but it is desirable. It is out of the if block. This part would be execute only in case of 
```
if  (r.Retries < Max){
     if (we find 200, 404, 404 ){
         execute delete;
}
} else {
        execute delete;
}
```
The actual logic is written in a different way

internal/proxy/proxy.go
kishansagathiya: It is already time.Duration

internal/proxy/github.go
kishansagathiya: :+1:

internal/proxy/github.go
kishansagathiya: :+1:

internal/proxy/github.go
kishansagathiya: no

internal/proxy/github.go
sthaha: > ns and clusterURL are being logged only once. 
Actually no, look below for 
```go
log.WithField("ns", ns).
```

This information is useful while debugging as the inspecting the namespace usually leads to more clues.

internal/proxy/proxy.go
sthaha: thanks, `time.Sleep(p.bufferCheckSleep * time.Second)` in the previous code confused me.

internal/proxy/github.go
```diff
@@ -175,23 +177,25 @@ func (p *Proxy) ProcessBuffer() {
 								break
 							}
 
-							if resp.StatusCode != 200 && resp.StatusCode != 404 { //Retry later if the response is not 200
-								log.Error(fmt.Sprintf("Got status %s after retrying request on %s", resp.Status, req.URL))
+							if resp.StatusCode == 200 {
+								log.WithField("ns", ns).Infof("Request to %q forwarded.", req.Host)
+							} else if resp.StatusCode == 404 || resp.StatusCode == 400 {
+								log.Warn("Got status %s after retrying request on %s, throwing away the request", resp.Status, req.URL)
```
sthaha: did you mean Warnf instead?

internal/proxy/github.go
```diff
@@ -175,23 +177,25 @@ func (p *Proxy) ProcessBuffer() {
...
+							} else {
+								//Retry later if the response is not 200 or 400 or 404
+								log.Error("Got status %s after retrying request on %s", resp.Status, req.URL)
```
sthaha: Errorf?

internal/proxy/github.go
kishansagathiya: ok

internal/proxy/github.go
kishansagathiya: Ok

internal/proxy/github.go
```diff
@@ -89,7 +91,7 @@ func (p *Proxy) handleGitHubRequest(w http.ResponseWriter, r *http.Request, requ
 	okToForward = true
 	r.Body = ioutil.NopCloser(bytes.NewReader(body))
 	//If Jenkins is up, we can simply proxy through
-	requestLogEntry.WithField("ns", ns).Infof(fmt.Sprintf("Passing through %s", r.URL.String()))
+	nsLogger.Infof(fmt.Sprintf("Passing through %s", r.URL.String()))
```
sthaha: with Infof we don't need fmt.Sprintf as mentioned earlier

internal/proxy/github.go
```diff
@@ -175,23 +178,25 @@ func (p *Proxy) ProcessBuffer() {
 								break
 							}
 
-							if resp.StatusCode != 200 && resp.StatusCode != 404 { //Retry later if the response is not 200
-								log.Error(fmt.Sprintf("Got status %s after retrying request on %s", resp.Status, req.URL))
+							if resp.StatusCode == 200 {
+								log.WithField("ns", ns).Infof("Request to %q forwarded.", req.Host)
```
sthaha: nsLogger here

internal/proxy/github.go
```diff
@@ -175,23 +178,25 @@ func (p *Proxy) ProcessBuffer() {
...
+							} else if resp.StatusCode == 404 || resp.StatusCode == 400 {
+								log.Warnf("Got status %s after retrying request on %s, throwing away the request", resp.Status, req.URL)
```
sthaha: same, nsLogger here as well

internal/proxy/github.go
```diff
@@ -53,13 +53,15 @@ func (p *Proxy) handleGitHubRequest(w http.ResponseWriter, r *http.Request, requ
...
-	clusterURL := namespace.ClusterURL
-	requestLogEntry.WithFields(log.Fields{"ns": ns, "cluster": clusterURL, "repository": gh.Repository.CloneURL}).Info("Processing GitHub request ")
 	if err != nil {
 		p.HandleError(w, err, requestLogEntry)
 		return
 	}
+	ns = namespace.Name
+	clusterURL := namespace.ClusterURL
+	nsLogger := requestLogEntry.WithField("ns", ns)
+	nsLogger.WithFields(log.Fields{"cluster": clusterURL, "repository": gh.Repository.CloneURL}).Info("Processing GitHub request ")
```
sthaha: trivial: can use `WithField` instead

internal/proxy/github.go
```diff
@@ -175,23 +180,25 @@ func (p *Proxy) ProcessBuffer() {
...
 
-							if resp.StatusCode != 200 && resp.StatusCode != 404 { //Retry later if the response is not 200
-								log.Error(fmt.Sprintf("Got status %s after retrying request on %s", resp.Status, req.URL))
+							if resp.StatusCode == 200 {
+								nsLogger.Infof("Request to %q forwarded.", req.Host)
+							} else if resp.StatusCode == 404 || resp.StatusCode == 400 {
+								log.Warnf("Got status %s after retrying request on %s, throwing away the request", resp.Status, req.URL)
+							} else {
+								//Retry later if the response is not 200 or 400 or 404
+								log.Errorf("Got status %s after retrying request on %s", resp.Status, req.URL)
```
sthaha: must be `req.URL.String()`

PR #296, opened 2018-06-08
Fix #295 Scale down proxy to 1

### LINKED ISSUE #295: Scale down proxy to 1
We have seen some issues with proxy, as described by @sthaha  : 

```
@aslak do you know if the openshift LB when we scale the pods  would inspect the requests for cookies  and direct it the same pod that handled the request earlier? 
I am asking this because we have scaled the jenkins-proxy to 3 pods and the state if the user is authenticated is maintained in memory of each pod and if the proxy finds if the cookie isn't in the in-memory cache, it would delete the cookie, forcing a re-authentication.
```

Let scale down the proxy to 1 until we figured it out

chmouel:


alien-ike:
### Ike Plugins (test-keeper)

Thank you @chmouel for this contribution!



alien-ike:
Hey @chmouel! It seems you tried to trigger `/ok-without-tests` command, but this will not have any effect due to insufficient permission. You have to be admin or requested reviewer or pull request approver, but not pull request creator for this command to take an effect.

PR #300, opened 2018-06-13
Issue #299 Enable CORS properly

### LINKED ISSUE #299: CORS is not properly enabled
We have only set allowed credentials and have not set the AllowedOrigins 
Look at https://github.com/rs/cors#parameters 
https://chat.openshift.io/developers/pl/dze3unuwpbgwmfr7eycc9r6mtw

kishansagathiya:
Set AllowOrigins and AllowMethods
Removed unnecessary CORS support for unexposed API and removed
their tests
Added tests for CORS support of exposed Jenkins API
Fixes #299

### DIFF
--- a/cmd/fabric8-jenkins-proxy/main.go
+++ b/cmd/fabric8-jenkins-proxy/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"net/http"
 	"os"
-	"strings"
 
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/api"
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/clients"
@@ -263,10 +262,10 @@ func newAPIServer(api api.ProxyAPI) *http.Server {
 }
 
 func newJenkinsAPIServer(jenkinsAPI jenkinsapi.JenkinsAPI, config configuration.Configuration) *http.Server {
-	allowedMethods := strings.Split(config.GetAllowedOrigins(), ", ")
+	allowedOrigins := config.GetAllowedOrigins()
 	c := cors.New(cors.Options{
 		AllowCredentials: true,
-		AllowedOrigins:   allowedMethods,
+		AllowedOrigins:   allowedOrigins,
 		AllowedMethods:   []string{"POST"},
 	})
 	srv := &http.Server{
--- a/cmd/fabric8-jenkins-proxy/main.go
+++ b/cmd/fabric8-jenkins-proxy/main.go
@@ -255,19 +255,17 @@ func setupSignalChannel(cancel context.CancelFunc) {
 }
 
 func newAPIServer(api api.ProxyAPI) *http.Server {
-	c := cors.New(cors.Options{
-		AllowCredentials: true,
-	})
-	srv := &http.Server{
+	return &http.Server{
 		Addr:    apiRouterPort,
-		Handler: c.Handler(router.CreateAPIRouter(api)),
+		Handler: router.CreateAPIRouter(api),
 	}
-	return srv
 }
 
 func newJenkinsAPIServer(jenkinsAPI jenkinsapi.JenkinsAPI) *http.Server {
 	c := cors.New(cors.Options{
 		AllowCredentials: true,
+		AllowedOrigins:   []string{"https://openshift.io", "https://*.openshift.io"},
+		AllowedMethods:   []string{"POST"},
 	})
 	srv := &http.Server{
 		Addr:    jenkinsAPIRouterPort,
--- a/cmd/fabric8-jenkins-proxy/main.go
+++ b/cmd/fabric8-jenkins-proxy/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/api"
 	"github.com/fabric8-services/fabric8-jenkins-proxy/internal/clients"
@@ -174,7 +175,7 @@ func startWorkers(
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		srv := newJenkinsAPIServer(jenkinsAPI)
+		srv := newJenkinsAPIServer(jenkinsAPI, config)
 
 		go func() {
 			mainLogger.Infof("Starting Jenkins Status API router on port %s", jenkinsAPIRouterPort)
@@ -261,10 +262,11 @@ func newAPIServer(api api.ProxyAPI) *http.Server {
 	}
 }
 
-func newJenkinsAPIServer(jenkinsAPI jenkinsapi.JenkinsAPI) *http.Server {
+func newJenkinsAPIServer(jenkinsAPI jenkinsapi.JenkinsAPI, config configuration.Configuration) *http.Server {
+	allowedMethods := strings.Split(config.GetAllowedOrigins(), ", ")
 	c := cors.New(cors.Options{
 		AllowCredentials: true,
-		AllowedOrigins:   []string{"https://openshift.io", "https://*.openshift.io"},
+		AllowedOrigins:   allowedMethods,
 		AllowedMethods:   []string{"POST"},
 	})
 	srv := &http.Server{
--- a/cmd/fabric8-jenkins-proxy/main_cors_test.go
+++ b/cmd/fabric8-jenkins-proxy/main_cors_test.go
@@ -29,7 +29,7 @@ func TestAPIServerCORSHeaders(t *testing.T) {
 	config := mock.NewConfig()
 	apiServer := newJenkinsAPIServer(&MockJenkinsAPIImpl{}, &config)
 
-	reader, _ := http.NewRequest("POST", "/jenkins/start", nil)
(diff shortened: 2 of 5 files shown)


invincibleJai:
@kishansagathiya we need to allow localhost as well as this would be useful for testing/dev work while working with fabric8-ui

cmd/fabric8-jenkins-proxy/main.go
```diff
@@ -255,19 +255,17 @@ func setupSignalChannel(cancel context.CancelFunc) {
...
-		Handler: c.Handler(router.CreateAPIRouter(api)),
+		Handler: router.CreateAPIRouter(api),
 	}
-	return srv
 }
 
 func newJenkinsAPIServer(jenkinsAPI jenkinsapi.JenkinsAPI) *http.Server {
 	c := cors.New(cors.Options{
 		AllowCredentials: true,
+		AllowedOrigins:   []string{"https://openshift.io", "https://*.openshift.io"},
```
sthaha: would this cover jenkins.prod-preview.openshift.io?

sthaha: How about making this configurable?

cmd/fabric8-jenkins-proxy/main_cors_test.go
```diff
@@ -6,35 +6,39 @@ import (
...
 func TestAPIServerCORSHeaders(t *testing.T) {
-	apiServer := newAPIServer(MockProxyAPIImpl{})
-	reader, _ := http.NewRequest("GET", "/", nil)
+	apiServer := newJenkinsAPIServer(&MockJenkinsAPIImpl{})
+	reader, _ := http.NewRequest("POST", "/jenkins/start", nil)
+	// Check for origin "https://*.openshift.io"
 	randomOrigin := uuid.NewV4().String()
-	reader.Header.Set("origin", randomOrigin) // allowing everything for now.
-
+	reader.Header.Set("origin", "https://"+randomOrigin+".openshift.io")
```
sthaha: would need to test jenkins.prod-preview.openshift.io

cmd/fabric8-jenkins-proxy/main_cors_test.go
```diff
@@ -6,35 +6,39 @@ import (
...
-	assert.Equal(t, randomOrigin, writer.Header().Get("access-control-allow-origin"))
-
+	// Check for origin "https://openshift.io"
+	reader.Header.Set("origin", "https://.openshift.io")
+	writer = httptest.NewRecorder()
+	apiServer.Handler.ServeHTTP(writer, reader)
+	assert.Equal(t, "https://.openshift.io", writer.Header().Get("access-control-allow-origin"))
 }
 
 func TestProxyCORSHeaders(t *testing.T) {
```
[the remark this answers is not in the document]
sthaha: and it passes?? I am confused as to what the test is doing then.

sthaha:
@invincibleJai 
> @kishansagathiya we need to allow localhost as well as this would be useful for testing/dev work while working with fabric8-ui

I am not sure if we want to allow this for a production system. You should be able to turn off CORS using a browser extension/settings.

cmd/fabric8-jenkins-proxy/main_cors_test.go
```diff
@@ -6,35 +6,39 @@ import (
...
+type MockJenkinsAPIImpl struct{}
 
-func (m MockProxyAPIImpl) Info(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
-	resp := api.Response{}
+func (api *MockJenkinsAPIImpl) Start(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	resp := clients.StatusResponse{}
 	json.NewEncoder(w).Encode(resp)
 }
 
 func TestAPIServerCORSHeaders(t *testing.T) {
```
sthaha: could you also add a negetive test for  non openshift.io origins

invincibleJai:
> I am not sure if we want to allow this for a production system. You should be able to turn off CORS using a browser extension/settings.

for all other OSIO services it is enabled I believe and it's just not me anyone working on fabric8-ui, would face issue in local dev.

kishansagathiya:
@invincibleJai If we can turn off CORS by changing browser settings or can be done manually, that would be the right this to do.
I found https://stackoverflow.com/questions/3102819/disable-same-origin-policy-in-chrome. Try to turn off CORS manually. If we can't do it, we would add it.

invincibleJai:
@kishansagathiya you are right it can be done, my point is for none of the services in OSIO we need to do that and developers need to be informed about it as almost many of us run fabric8-ui in local with sourcing to prod/prod-preview

kishansagathiya:
@invincibleJai 
I am making origins configurable, so we would be able to change it whenever we want.

cmd/fabric8-jenkins-proxy/main_cors_test.go
sthaha: is that because randomOrigin will contain one or more `.` in between?

cmd/fabric8-jenkins-proxy/main.go
```diff
@@ -255,19 +256,18 @@ func setupSignalChannel(cancel context.CancelFunc) {
...
 		Addr:    apiRouterPort,
-		Handler: c.Handler(router.CreateAPIRouter(api)),
+		Handler: router.CreateAPIRouter(api),
 	}
-	return srv
 }
 
-func newJenkinsAPIServer(jenkinsAPI jenkinsapi.JenkinsAPI) *http.Server {
+func newJenkinsAPIServer(jenkinsAPI jenkinsapi.JenkinsAPI, config configuration.Configuration) *http.Server {
+	allowedMethods := strings.Split(config.GetAllowedOrigins(), ", ")
```
sthaha: rather let the config.GetAllowedOrigins return []string{} so that the caller need not worry about the delimiter. also using a single char `,` as delimiter will be a better choice.

sthaha: Another thing that is that the var name shouldn't be `allowedMethods` :)

cmd/fabric8-jenkins-proxy/main_cors_test.go
```diff
@@ -6,46 +6,47 @@ import (
...
-	apiServer := newAPIServer(MockProxyAPIImpl{})
-	reader, _ := http.NewRequest("GET", "/", nil)
-	randomOrigin := uuid.NewV4().String()
-	reader.Header.Set("origin", randomOrigin) // allowing everything for now.
+	config := mock.NewConfig()
+	apiServer := newJenkinsAPIServer(&MockJenkinsAPIImpl{}, &config)
 
+	reader, _ := http.NewRequest("POST", "/jenkins/start", nil)
```
sthaha: isn't the endpoint `/api/jenkins/start` ?

internal/configuration/env_config.go
```diff
@@ -57,6 +58,7 @@ func init() {
 	settings["GetDebugMode"] = Setting{"JC_DEBUG_MODE", defaultDebugMode, []func(interface{}, string) error{util.IsBool}}
 	settings["GetHTTPSEnabled"] = Setting{"JC_ENABLE_HTTPS", defaultHTTPSEnabled, []func(interface{}, string) error{util.IsBool}}
 	settings["GetGatewayTimeout"] = Setting{"JC_GATEWAY_TIMEOUT", defaultGatewayTimeout, []func(interface{}, string) error{util.IsDuration}}
+	settings["GetAllowedOrigins"] = Setting{"JC_ALLOWED_ORIGINS", defaultAllowedOrigins, []func(interface{}, string) error{util.IsNotEmpty}}
```
sthaha: defaultAllowedOrigins should be `[]string` instead so that the caller need not worry about the implementation detail (how the string is delimited)

internal/configuration/env_config.go
```diff
@@ -23,6 +23,7 @@ const (
 	defaultDebugMode                 = "false"
 	defaultHTTPSEnabled              = "false"
 	defaultGatewayTimeout            = "25s"
+	defaultAllowedOrigins            = "https://*openshift.io, https://localhost:*"
```
sthaha: should this be `*.openshift.io` instead? ?

kishansagathiya: This will cover xyz.openshift.io and openshift.io both. Otherwise I will have to put them separately. Like
```
defaultAllowedOrigins            = "https://openshift.io, https://*.openshift.io, https://localhost:*"
```

cmd/fabric8-jenkins-proxy/main_cors_test.go
kishansagathiya: You are right. This has to change. The test will pass because we are not testing the API. We are merely testing if this server allows external requests. So, the endpoint here will not matter.

internal/configuration/env_config.go
kishansagathiya: Not sure, how this can be done. I don't see much value here anyway. I would rather skip this. If you want this feel free to modify this PR.

cmd/fabric8-jenkins-proxy/main.go
kishansagathiya: Ack

Issue #303, opened 2018-06-15
Provide an api to get the status of jenkins pods 


sthaha:
Currently we provide an api to start jenkins `api/jenkins/start` which returns the status of the pods
But there are usecases UI only wants to know the state of the pod before calling the start api.
So lets provide an api to get the status of jenkins pods `GET` `/api/jenkins/status` that returns
the status similar to the `start` api.

PR #310, opened 2018-06-22
Issue #303 Provide an api to get the status of jenkins pods (#303)

rohanKanojia:
There are lots of use cases where UI only wants to know the state of pod
before calling the api. So lets provide an api to get the status of jenkins
pods GET /api/jenkins/status that returns status like start api.

#303

### DIFF
--- a/internal/jenkinsapi/jenkinsapi.go
+++ b/internal/jenkinsapi/jenkinsapi.go
@@ -15,6 +15,7 @@ import (
 //JenkinsAPI contains API to check whether Jenkins for the current user is idle|running|starting
 type JenkinsAPI interface {
 	Start(w http.ResponseWriter, r *http.Request, ps httprouter.Params)
+	Status(w http.ResponseWriter, r *http.Request, ps httprouter.Params)
 }
 
 // JenkinsAPIImpl implements JenkinsAPI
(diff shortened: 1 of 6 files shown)


hrishin:
@sthaha review, please.

internal/jenkinsapi/jenkinsapi.go
```diff
@@ -31,6 +32,34 @@ func NewJenkinsAPI(tenant clients.TenantService, idler clients.IdlerService) Jen
...
+// Returns the Jenkins pods status for current user
+func (api *jenkinsAPIImpl) Status(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	resp := clients.StatusResponse{}
+
+	authHeader := r.Header.Get("Authorization")
+	if !strings.HasPrefix(authHeader, "Bearer ") {
+		HandleError(w, resp, errors.New("Could not find Bearer token in Authorization Header"), http.StatusUnauthorized)
+		return
+	}
+	accessToken := strings.Split(r.Header.Get("Authorization"), " ")[1]
```
kishansagathiya: accessToken := strings.Split(authHeader, " ")[1]

sthaha: The token parsing part is repetitive, this must be moved to a function.

internal/jenkinsapi/jenkinsapi.go
```diff
@@ -31,6 +32,34 @@ func NewJenkinsAPI(tenant clients.TenantService, idler clients.IdlerService) Jen
...
+	namespace, err := api.tenant.GetNamespace(accessToken)
+	if err != nil {
+		HandleError(w, resp, err, http.StatusUnauthorized)
+		return
+	}
+	log.Infof("Found token info in the query. Namespace is %s and clusterURL is %s", namespace.Name, namespace.ClusterURL)
+
+	status, err := api.idler.State(namespace.Name, namespace.ClusterURL)
```
sthaha: this part is also the same as in start. Please move to a different function.

internal/jenkinsapi/jenkinsapi.go
```diff
@@ -31,32 +32,55 @@ func NewJenkinsAPI(tenant clients.TenantService, idler clients.IdlerService) Jen
...
+
+	return status, err
+}
+
+// Returns the Jenkins pods status for current user
+func (api *jenkinsAPIImpl) Status(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	resp := clients.StatusResponse{}
+
+	namespace, _ := fetchNamespace(w, r, api.tenant)
+	status, _ := fetchState(w, r, namespace, api.idler)
```
sthaha: I am not sure if need to refactor out fetchState since it doesn't seem to save us anything if we move the error handling to the caller.

internal/jenkinsapi/jenkinsapi.go
```diff
@@ -31,32 +32,55 @@ func NewJenkinsAPI(tenant clients.TenantService, idler clients.IdlerService) Jen
...
-		return
+		return namespace, err
 	}
 	log.Infof("Found token info in the query. Namespace is %s and clusterURL is %s", namespace.Name, namespace.ClusterURL)
+	return namespace, nil
+}
+
+// Queries Idler state and returns it
+func fetchState(w http.ResponseWriter, r *http.Request, namespace clients.Namespace, idler clients.IdlerService) (clients.PodState, error) {
+	resp := clients.StatusResponse{}
```
sthaha: resp is only used within the if; move the declaration close to the usage.

internal/jenkinsapi/jenkinsapi.go
```diff
@@ -31,32 +32,55 @@ func NewJenkinsAPI(tenant clients.TenantService, idler clients.IdlerService) Jen
...
 		HandleError(w, resp, err, http.StatusUnauthorized)
-		return
+		return namespace, err
 	}
 	log.Infof("Found token info in the query. Namespace is %s and clusterURL is %s", namespace.Name, namespace.ClusterURL)
+	return namespace, nil
+}
+
+// Queries Idler state and returns it
+func fetchState(w http.ResponseWriter, r *http.Request, namespace clients.Namespace, idler clients.IdlerService) (clients.PodState, error) {
```
sthaha: this method seems unnecessary as to "fetch" the state of the pod. it seesm to require not only the namespace  details but also a http request and a response.

internal/jenkinsapi/jenkinsapi.go
```diff
@@ -31,32 +32,55 @@ func NewJenkinsAPI(tenant clients.TenantService, idler clients.IdlerService) Jen
...
+	resp := clients.StatusResponse{}
 
-	status, err := api.idler.State(namespace.Name, namespace.ClusterURL)
+	status, err := idler.State(namespace.Name, namespace.ClusterURL)
```
sthaha: if we move the err handling to the caller this function boils down to just 
`return idler.State(namespace.Name, namespace.ClusterURL)` hence we shouldn't need this function.

internal/router/router_test.go
```diff
@@ -90,3 +106,13 @@ func Test_JenkinsAPI_routes_are_setup(t *testing.T) {
 	mockedRouter.ServeHTTP(w, req)
 	require.Equal(t, "{\"data\":{\"state\":\"\"}}\n", w.GetBody(), "Routing failed for /api/jenkins/start")
 }
+
+func Test_JenkinsAPI_status_is_fetched(t *testing.T) {
+	mockedJenkinsAPI := &mockJenkinsAPI{}
```
sthaha: is this method testing anything more than the **mock** jenkins api?

rohanKanojia: Yes, I think it's only testing `MockJenkinsAPI` . But I saw a similar implementation for `/api/jenkins/start` so I wrote it. I can remove it if you don't think it's required.

PR #316, opened 2018-09-03
Remove dependency on KeyCloak

sthaha:
Removes dependency on Keycloak. See commit messages for details.


internal/auth/auth.go
```diff
@@ -0,0 +1,241 @@
...
+			err = fmt.Errorf("Could not find user id in token")
+			return
+		}
+		sub = claims["sub"].(string)
+	}
+	return
+}
+
+// OSOTokenForCluster returns Openshift online token given the clusterURL and raw JWT token
+func (c *Client) OSOTokenForCluster(clusterURL, token string) (osoToken string, err error) {
```
kishansagathiya: Shouldn't `token` be `accessToken`? Can you confirm that we are using raw JWT token here?

internal/auth/auth_test.go
```diff
@@ -0,0 +1,160 @@
...
+    },
+    {
+      "key": "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA0eiOo86RtUEil7vbUQWW2rXDSNnGqXjXxxSAeCoBv5ecWGH7ZagmoFPfjSfjxT2qdQcTlRMkQzwsrztPbl3nKYkj/SCgxk0BPVQKWKJObe2bYIzxK8a+s2yvcS7LNrKV5+oWrOG6X46KQJbNh3SyIGy1o0lbBiOTNkQl984JgL9iuxJByz5azjU5W36IP6Hi2eJEYZxmve+I6jGF7PGx6sa7vNHcRDdxBU578IFw5CaMq+2zfza2AQwEH4LKb2bnTmbotgYPjhANgSl6oiqWehzFqeyDuZGJZAzkj31NLguNhmd6MbKCx5QmGJB7907cN1slvYcRA8UTKImfaK3+wQIDAQAB",
+      "kid": "PE6-BEECZZpPZIVxLR6NinbthOHJcGqYrfl8v7v6BVA"
+    }
+  ]
+}
+`
+
+type fakeAuthService struct {
```
kishansagathiya: mockAuthService?

internal/auth/auth.go
sthaha: no, it is expecting the raw token .. see below 

```go
	if len(tj.AccessToken) > 0 {
		osoToken = tj.AccessToken
	} else {
		err = fmt.Errorf("OSO access token empty for %s", c.URL)
	}
```

internal/auth/auth_test.go
sthaha: fakeAuth is equally unambiguous but if I have to make changes to the code to address any comments, I will change it mock.

internal/auth/auth.go
sthaha: @kishansagathiya no, you are right, it must be called AccessToken
so, I will change it to `accessToken`

kishansagathiya:
While reviewing this, https://github.com/fabric8-services/fabric8-jenkins-proxy/issues/317 occurred

Gopkg.toml
```diff
@@ -44,3 +44,7 @@
 [[constraint]]
   name = "gopkg.in/ory-am/dockertest.v3"
   version = "=v3.1.3"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/matryer/resync"
```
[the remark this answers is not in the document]
sthaha: @chmouel by drop the author meant - https://github.com/matryer/drop which I am not so inclined to.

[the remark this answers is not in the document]
chmouel: it's a 40 lines file  but yeah not a big issue,

PR #319, opened 2018-09-04
Fix #4083 Lower initial wait for Jenkins to start

sthaha:
With more resources allocated to Jenkins, it usually start under
30 seconds, so this patch reduces the initial waiting time to 45 seconds


alien-ike:
### Ike Plugins (test-keeper)

For more information please head over to official [documentation](http://arquillian.org/ike-prow-plugins/#_test_keeper_plugin). You can find there how to configure the plugin.

chmouel:
Maybe let's take an opportunity to make it as a const somwhere?

PR #334, opened 2018-10-24
Handle errors instead of redirecting

### LINKED ISSUE #330: Proxy should be return errors or NotFounds on error not retrying again and again
See https://github.com/openshiftio/openshift.io/issues/4180

chmouel:
When there is an errors instead of redirecting forever, just show it and exit.

This avoid the catch all issues "maximum redirect" and show a real error for
users to use instead.

This closes #330 and openshiftio/openshift.io#4180


alien-ike:
### Ike Plugins (test-keeper)

Thank you @chmouel for this contribution!



Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/df49d2c83bc938c5e7eb5b1236b4fdcf15b25849/.ike-prow/test-keeper.yaml).

PR #342, opened 2018-12-03
Webhook from github to jenkins fails with 500 Status

kishansagathiya:
The code was using an empty namespace, fixed that

Fixes https://github.com/openshiftio/openshift.io/issues/4600

### DIFF
--- a/internal/proxy/github.go
+++ b/internal/proxy/github.go
@@ -103,6 +103,7 @@ func (p *Proxy) handleGitHubRequest(w http.ResponseWriter, r *http.Request, requ
 	r.Body = ioutil.NopCloser(bytes.NewReader(body))
 	//If Jenkins is up, we can simply proxy through
 	nsLogger.Infof("Passing through %s", r.URL.String())
+	w.WriteHeader(http.StatusOK)
 	return
 }
 
--- a/internal/proxy/github.go
+++ b/internal/proxy/github.go
@@ -57,6 +57,8 @@ func (p *Proxy) handleGitHubRequest(w http.ResponseWriter, r *http.Request, requ
 		return
 	}
 
+	ns = namespace.Name
+
 	nsLogger := requestLogEntry.WithField("ns", ns)
 
 	pci := CacheItem{
--- a/internal/proxy/github_test.go
+++ b/internal/proxy/github_test.go
@@ -6,18 +6,71 @@ import (
 	"net/http/httptest"
 	"testing"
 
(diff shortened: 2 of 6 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/d6a72d7bebfe7d4528296ff98af2b6f343cbeda9/.ike-prow/test-keeper.yaml).

internal/proxy/github.go
```diff
@@ -57,6 +57,8 @@ func (p *Proxy) handleGitHubRequest(w http.ResponseWriter, r *http.Request, requ
 		return
 	}
 
+	ns = namespace.Name
```
hrishin: @kishansagathiya do we have any test case for this in place?

kishansagathiya: There are some tests for GitHub webhook requests, but since this mistake was not detected by test, clearly we can do with better or more tests.

khrm: Can you change https://github.com/fabric8-services/fabric8-jenkins-proxy/blob/140b669a04d1a98a9c75d3161f3822fa34a88f83/internal/proxy/github_test.go#L32  to check for return values?

internal/proxy/github_test.go
```diff
@@ -29,7 +29,11 @@ func TestGHWebHookRequest(t *testing.T) {
 	w := httptest.NewRecorder()
 	req := httptest.NewRequest("GET", "http://proxy", bytes.NewBuffer(gh))
 	req.Header.Add("User-Agent", "GitHub-Hookshot"+uuid.NewV4().String())
-	p.handleGitHubRequest(w, req, proxyLogger)
+	ns, okToForward := p.handleGitHubRequest(w, req, proxyLogger)
+
+	assert.Equal(t, ns, "namespace-jenkins")
+	assert.False(t, okToForward, "It should not be ok to forward")
```
chmouel: You are testing the "good path" can you handle the "bad path" too ?

kishansagathiya: sure, I will think of some bad paths and add them

kishansagathiya:
[ test ]

piyush-garg:
[retest]

piyush-garg:
[test]

internal/proxy/github.go
kishansagathiya: This added in commits that came after this comment.

piyush-garg:
Merging this.

